PR #27, opened 2022-12-30
Csv sep for bulk

shostetter:
Fixes bug for importing large CSVs with non-comma delimited values


chriscah:
hate to add to PR-ception but: https://github.com/safety-analytics-mapping/ris/pull/228 any chance this is working for you now? If so, I can put a PR up for pysqldb3 too and we can just use ogr? Or we can go this route for now if you prefer

shostetter:
@chriscah https://github.com/safety-analytics-mapping/ris/pull/228 isnt throwing an error but i think the quote strings arent working correctly. Tried exporting all of the 2022 nypd crashes and some records are messed up in the output

chriscah (approved):

chriscah:
Can you send me an example of the error? We can merge this in for the time being but want to switch over eventually

PR #34, opened 2023-02-15
Import shp from zip

shostetter:
Adds zip Boolean flag to pysqldb3.DbConnect.shp_to_table and shapefile.read_shp  to allow reading shape files from zipped directories


chriscah (approved):
looks good!

PR #36, opened 2023-02-16
OGR Query to CSV 

chriscah:
Changes to use ogr for query/table to CSV 

In tandem with: https://github.com/safety-analytics-mapping/ris/pull/228/files


shostetter (approved):

chriscah:
adding some changes to tests, should (more or less) match https://github.com/safety-analytics-mapping/ris/pull/228/files

PR #39, opened 2023-03-14
CSV / XLS to table kwargs to pass to pandas 

shostetter:
- Allows you to pass pandas keyword arguments for csv/xls to table to take advantage of skiprows etc.
- Allows query_to_csv to work with `"` encapsulated column names in PG, needed for columns with special characters

### DIFF
--- a/pysqldb3/pysqldb3.py
+++ b/pysqldb3/pysqldb3.py
@@ -9,6 +9,7 @@ import plotly.express as px
 import configparser
 import os
 from .Config import write_config
+import pyarrow.csv as pyarrowcsv
 
 write_config(confi_path=os.path.dirname(os.path.abspath(__file__)) + "\\config.cfg")
 config = configparser.ConfigParser()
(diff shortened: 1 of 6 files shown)


shostetter:
I believe this works though was getting some failures in unrelated areas, rerunning sometimes fixed them so not sure whats going on

pysqldb3/pysqldb3.py
```diff
@@ -1195,133 +1207,36 @@ def xls_to_table(self, input_file=None, sheet_name=0, overwrite=False, schema=No
...
-                # Match previous styles
-                cols = []
-                for c in sheet_data[0]:
-                    if type(c) == str:
-                        cols.append(c.strip().replace(' ', '_').replace('.', '_'))
-                    else:
-                        cols.append(c)
 
-                df = pd.DataFrame(sheet_data[1:], columns=cols)
+        try:
```
chriscah, line 1220: We can get rid of most of the code here but maybe large tables should still default to _bulk_xlsx_to_table

chriscah (commented):
a few small comments

pysqldb3/pysqldb3.py
```diff
@@ -1228,15 +1228,37 @@ def xls_to_table(self, input_file=None, sheet_name=0, overwrite=False, schema=No
...
+                    df.to_csv(temp_file, index=False, header=True)
+
+                    success = self._bulk_csv_to_table(input_file=temp_file, schema=schema, table=table,
+                                                      table_schema=table_schema, days=days)
+                    os.remove(temp_file)
+
+                    if not success:
+                        raise AssertionError('Bulk file loading failed.'.format(schema, table))
+
+                except SystemExit:
```
chriscah, line 1251: if os.remove doesnt error if the file doesnt exist, could always try to remove in except as well if a problem happens midflight

chriscah (approved):
most recent changes look good!

PR #40, opened 2023-03-30
Pyarrow and integer fixes for pandas imports 

shostetter:
This is a fix for an issue when bulk importing csv files. Some field will be typed as integers but when bulk importing gdal includes a decimal place which then breaks the casting to int on the insert into the final table. This resolves that, though I'm sure there is a more elegant way to resolve this

### DIFF
--- a/pysqldb3/pysqldb3.py
+++ b/pysqldb3/pysqldb3.py
@@ -807,6 +807,10 @@ class DbConnect:
         for col in df.dtypes.items():
             col_name, col_type = col[0], type_decoder(col[1], varchar_length=allowed_length)
 
+            # autodetect date and force to text (common error)
+            if 'date' in col_name.lower() and col_type in ('int', 'bigint', 'float'):
+                col_type = 'varchar(500)'
+
             if column_type_overrides and col_name in column_type_overrides.keys():
                 input_schema.append([clean_column(col_name), column_type_overrides[col_name]])
             else:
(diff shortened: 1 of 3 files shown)


pysqldb3/pysqldb3.py
```diff
@@ -807,6 +807,10 @@ def dataframe_to_table_schema(self, df, table, schema=None, overwrite=False, tem
         for col in df.dtypes.items():
             col_name, col_type = col[0], type_decoder(col[1], varchar_length=allowed_length)
 
+            # autodetect date and force to text (common error)
+            if 'date' in col_name.lower() and col_type in ('int', 'bigint', 'float'):
```
Horaceshi624, line 811: Not sure about this but Do all the columns with date type data in ris have 'date' in their name?

shostetter: no, it could miss some, but it was missing some date fields and I figured this was a relatively safe assumption, since worse case it stores as text which will prevent errors but not introduce any new ones. If it tries as int and finds a date it will fail, but if it assumes it to be text it will always work

pysqldb3/tests/test_csv_to_table.py
```diff
@@ -341,6 +415,79 @@ def test_csv_to_table_basic(self):
...
+        csv_df.columns = [c.replace(' ', '_').lower() for c in list(csv_df.columns)]  # TODO: look into this difference
+        csv_df = csv_df.astype({'false_date': 'int'})
+
+        # Assert df equality, excluding dtypes and columns
+        pd.testing.assert_frame_equal(db_df, csv_df, check_dtype=False)
+        # assert that dates have been set to varchar
+        for (name, typ) in sql.get_table_columns(create_table_name, schema=sql_schema):
+            if 'date' in name.lower() and name.lower() != 'false_date':
+                print(name, typ)
+                assert typ == 'varchar'
```
Horaceshi624, line 486: is this same as typ == 'character varying'?
And I think this should be good if there's no AssertionError raised.

shostetter:
Same thing but pg and Ms return different ways of describing it

Horaceshi624 (approved):
Hi Seth looks good and after test I didn't get any decimal place for int. thanks,

chriscah (approved):
looks good to me...this will be useful

PR #43, opened 2023-04-18
Connecting to Azure database with DbConnect

shostetter:
This is working for the basic functions, but we dont have a database we can write to so we are limited in what can be tested. 
Todo:
- temp tables are not allowed in the serverless pool,, so maybe a warning should be added if allow_temp_tables is passed with type azure 
- explore data_io features from Azure


chriscah:
@shostetter wow this was not as bad as I would have thought 👍

PR #48, opened 2023-09-28
Add 2 lines about tds_version to the parameters for MSS connectivity #47

cindyatnycdot:
Since I encountered a problem connecting to the MS SQL Server, I had to add an extra parameter where tds_version = r'7.0' in order for pymssql to work. In turn (since pysqldb3 is based on pymssql), I have to add this parameter to pysqldb3 for the Microsoft Server connection.


shostetter (commented):
adding comment here for consistency 
add a variable to the config and use that to determine if this new tds_version param is passed when the connection is made

cindyatnycdot:
If I add a `tds_version` variable to the config, are you referring to the `config.cfg` file or the `config.py` file? H

shostetter:
> If I add a `tds_version` variable to the config, are you referring to the `config.cfg` file or the `config.py` file?

both really, the .cfg file stores the values that are ready by the code at runtime, the .py file creates that .cfg fine when you install the library

cindyatnycdot:
I hope this code change will work. I basically removed a line from another section of code that I think wasn't required there and hopefully that was the single source of all the testing issues.

I did not change the config file. I tried adding the parameter to the config and then extracting it in the pysqldb3.py file, but despite trying ~10 different things, I couldn't change the config's value back to a raw string that would work successfully in the connection, so I've kept it as a raw string in the file,

PR #56, opened 2024-03-15
Add regex to remove '--' comments before running

cindyatnycdot:
I tried to run `pytest` to see what else passes/fails, but I could not get it to run (even when I commented out my changes). Will have to debug my version further, but I thought I would create the PR as the edit is small and I thought straight forward


shostetter (commented):
please add some tests for this condition in test_io.py to ensure it is working as expected.

cindyatnycdot:
I revised some of the tests in `test_io.py` because they weren't working for me (for instance, I can't `CREATE TABLES` in Microsoft SQL Server, and I can't write to `public`, only to `working`).

But the tests passed with comments integrated into them

pysqldb3/tests/test_io.py
```diff
@@ -614,32 +637,39 @@ def test_pg_to_pg_qry_basic_table(self):
                                 password=config.get('SECOND_PG_DB', 'DB_PASSWORD'))
 
         # Assert pg table doesn't exist
-        db.drop_table(schema=db.default_schema, table=test_pg_to_pg_qry_table)
-        assert not db.table_exists(table=test_pg_to_pg_qry_table)
+        db.drop_table(schema='working', table=test_pg_to_pg_qry_table)
```
shostetter, line 640: please add as variables to top of test file and use as parameters 
`pg_schema = 'working'`
`sql_schema = 'dbo'`

pysqldb3/tests/test_io.py
```diff
@@ -241,159 +241,184 @@
...
+        # Add test_table
+        # sql.drop_table(schema='dbo', table=test_sql_to_pg_qry_table)
+        # sql.query("""
+        # create table dbo.{} (test_col1 int, test_col2 int);
+        # insert into dbo.{} VALUES(1, 2);
+        # insert into dbo.{} VALUES(3, 4);
+        # """.format(test_sql_to_pg_qry_table, test_sql_to_pg_qry_table, test_sql_to_pg_qry_table))
+
+        # sql_to_pg_qry
+        data_io.sql_to_pg_qry(sql, db, query="""-- comments within the query
```
shostetter, line 259: if possible we shouldnt rely on tables being present for testing, you should be able to test with making a global temp table ie. `f'##{sql_table_name }`

pysqldb3/tests/test_io.py
```diff
@@ -241,159 +241,184 @@
...
+                                                select OFTCode, OnStreetName, FromStreetName, SegmentID
+                                                -- test
```
shostetter, line 261: i would doa simple test without comments 1st then add another test to validate the comments where handled correctly, it should include both `--` and `/*` type

shostetter (commented):
added a few comments to the test code

pysqldb3/tests/test_io.py
cindyatnycdot: I kept trying to get this to work on Fri/weekend, and it won't insert data into a global temp table when I run the `sql_to_pg_qry` function. I can create the temp table and `select` it, but the moment I add `insert` to the global temp table, it won't create the table in PG (it doesn't error out when I run the function, but if I try to access the table in PG, it doesn't exist).

Tried looking online to figure out the error, but can't find any existing literature.  Doesn't work for local temp table either. Not sure if you might have any luck with getting it to work on your end, or we need to change how we test this function without calling an existing table.

cindyatnycdot:
I replied to your comment on lines 256 - 259. The test for `sql_to_pg_qry` using temp tables currently doesn't work, and I've uploaded my latest code revisions to help us find a solution.

shostetter:
@cindyatnycdot all changes go into DEV, once there is enough change there we do a big update to MAIN

shostetter (approved):

PR #59, opened 2024-04-12
FIX - empty column datatype in csv import 

shostetter:
When importing data from csv (csv_to_table/bulk) if there are empty columns pandas defaults them to Int datatype, but often this creates issues when merging or updating if the field isnt really always null, but really just very sparsely populated. This identifies fields with this condition and sets the database datatype to varchar. 

Notes:
* Tests are included 
* pyarrow functions are not modified for this behavior

### DIFF
--- a/pysqldb3/tests/test_csv_to_table.py
+++ b/pysqldb3/tests/test_csv_to_table.py
@@ -53,7 +53,6 @@ class TestCsvToTablePG:
         # Cleanup
         db.drop_table(schema=pg_schema, table=create_table_name)
 
-
     def test_csv_to_table_basic_blank_col(self):
         # csv_to_table
         db.query('drop table if exists {}.{}'.format(pg_schema, create_table_name))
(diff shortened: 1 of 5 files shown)


pysqldb3/tests/test_csv_to_table.py
```diff
@@ -393,6 +417,59 @@ def test_bulk_csv_to_table_basic(self):
         # Cleanup
         db.drop_table(schema=pg_schema, table=create_table_name)
 
+    def test_bulk_csv_to_table_basic_blank_col(self):
+        # bulk_csv_to_table
+        db.query('drop table if exists {}.{}'.format(pg_schema, create_table_name))
+
+        fp = os.path.dirname(os.path.abspath(__file__)) + "\\test_data\\test8.csv"
```
cindyatnycdot, line 424: I saw that you use `DIR+` instead of `os.path.dirname(os.path.abspath(__file__))` in the `helpers.py` file. Is that something you wanted to change in this file too?

shostetter: yeah it should probably simplified to DIR

shostetter: resolved

Horaceshi624 (approved):

cindyatnycdot (approved):

Issue #53, opened 2024-03-11
pg_to_pg default permissions not being set correctly 


shostetter:
it should when creating a table set the permissions to grant select on [table] to public

PR #60, opened 2024-04-18
Call pysqldb3 to change permissions to public on pg_to_pg tables

cindyatnycdot:
This is related to the Issue #53.
Importing pysqldb3 was the only way that I could change permissions. Looked for a solution in ogr2ogr but couldn't find one.


cindyatnycdot:
I noticed that `data_io` is in python3, not python2, so calling the `pysqldb3.py` file within `data_io.py` should be compatible

shostetter (approved):
Looks good, all tests pass.

PR #68, opened 2024-08-09
Add initial functions for geopackage

cindyatnycdot:
Includes geopackage versions of all existing shapefile counterpart functions. Will also add functions to convert shapefile to geopackage and the reverse.


pysqldb3/geopackage.py
```diff
@@ -0,0 +1,500 @@
...
+        assert gpkg_name[-5:] == '.gpkg', "The input file should end with .gpkg . Please check your input."
+        assert shp_name[-4:] == '.shp', "The input file should end with .shp . Please check your input."
+
+        # same function as shape file
+        self.path, shp_name = parse_shp_path(self.path, shp_name)
+
+        if not self.path:
+            self.path = file_loc('folder')
+
+        self.cmd = WRITE_SHP_CMD_GPKG.format(gpkg_name = gpkg_name,
```
Horaceshi624, line 156: seems when writing table we don't use the param (tbl_name)? What is the usual value of self.table?

pysqldb3/pysqldb3.py
```diff
@@ -2073,3 +2074,216 @@ def feature_class_to_table(self, path, table, schema=None, shp_name=None, gdal_d
...
+        except Exception as e:
+            print(e)
+            pass
+
+        if self.type == PG:
+            self.query(f"""    
+            create temp table {tmp_table_name} as     
+            select * 
+            from ({query}) q 
+            limit 10
```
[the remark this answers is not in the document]
cindyatnycdot: How I understand what's happening (since this was copied from the `shp` function) is that it takes only 10 rows just to make the temp table small, and then it's using that temp table to parse out date/time columns so that it can reformat them before it runs the entire query

pysqldb3/cmds.py
```diff
@@ -76,6 +76,46 @@
...
+ogr2ogr --config GDAL_DATA "{gdal_data}" -nlt PROMOTE_TO_MULTI -overwrite -a_srs 
+"EPSG:{srid}" -f "PostgreSQL" PG:"host={host} user={user} dbname={dbname} 
+password={password}" "{gdb}" "{feature}" -nln {schema}.{tbl_name} -progress 
+""".replace('\n', ' ')
+
+READ_GPKG_FEATURE_CMD_MS = r"""
+ogr2ogr --config GDAL_DATA "{gdal_data}" -nlt PROMOTE_TO_MULTI -overwrite -a_srs 
+"EPSG:{srid}" -f MSSQLSpatial "MSSQL:server={ms_server};database={ms_db};UID={ms_user};PWD={ms_pass}"
+ "{gdb}" "{feature}" -nln {schema}.{tbl_name} -progress --config MSSQLSPATIAL_USE_GEOMETRY_COLUMNS NO {sf}
+""".replace('\n', ' ')
```
cindyatnycdot, line 113: I will need to write tests for this.. I included these commands but they weren't tested in the `Shapefile` tests that I copied

pysqldb3/pysqldb3.py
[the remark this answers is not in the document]
cindyatnycdot: But as an aside, Geopackage lets you name columns however long you want. This is different from shapefiles, which has a character limit

cindyatnycdot:
I looked into the additional functions I was going to write for geopackages, and realized that they don't apply to geopackages so I will delete. The commands were initially for converting ArcGIS geodatabase feature classes to Shapefile.

pysqldb3/cmds.py
Horaceshi624: So these tests are no longer needed, right?

cindyatnycdot: Yes, only these were removed

Horaceshi624 (approved):

pysqldb3/sql.py
```diff
@@ -204,3 +204,44 @@
...
+    FROM
+        sys.indexes AS a
+    INNER JOIN
+        sys.index_columns AS b
+        ON a.object_id = b.object_id AND a.index_id = b.index_id
+    WHERE
+        a.is_hypothetical = 0 
+        AND a.object_id = OBJECT_ID('{s}.{t}')
+        AND type_desc !='SPATIAL' -- taken care of by GDAL
+"""
```
shostetter, line 241: are these needed? any reason you cant use the `SHP_DEL_INDICES_QUERY_PG` / `SHP_DEL_INDICES_QUERY_MS`?

pysqldb3/pysqldb3.py
```diff
@@ -142,7 +143,7 @@ def __get_default_schema(self, db_type):
         """
         if db_type == MS:
             default_schema = self.dfquery('select schema_name()', internal=True)
-            default_schema = default_schema.iloc[0][0]#.encode('utf-8')
+            default_schema = default_schema.iloc[0, 0]#.encode('utf-8')
```
shostetter, line 146: this returns the same result, why is it needed?

pysqldb3/pysqldb3.py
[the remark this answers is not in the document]
shostetter: does geopackage allow datetime format? if so this is not needed we do it for shp since that datatype is not an option

pysqldb3/pysqldb3.py
```diff
@@ -2073,3 +2074,216 @@ def feature_class_to_table(self, path, table, schema=None, shp_name=None, gdal_d
...
+                                    col=col_name[2:-2], short_col=shortened_col)
+                elif self.type == MS:
+                    shortened_col = col_name[1:-1][:7]
+                    return_cols += " , cast([{col}] as date) [{short_col}_dt], cast(cast([{col}] as time) as varchar)" \
+                                   " [{short_col}_tm] ".format(
+                                    col=col_name[1:-1], short_col=shortened_col)
+
+        # Wrap the original query and select the non-datetime/timestamp columns and the parsed out dates/times
+        new_query = u"select {} from ({}) q ".format(return_cols, query)
+        Query.query_to_gpkg(self, new_query, path=path, gpkg_name=gpkg_name, cmd=cmd, gdal_data_loc=gdal_data_loc,
```
shostetter, line 2192: everything besides this call is the same a qry to shp, so you should extract the code that is the same into a sub/internal function that both can call and just pass in the call to different the Query methods

pysqldb3/pysqldb3.py
```diff
@@ -2073,3 +2074,216 @@ def feature_class_to_table(self, path, table, schema=None, shp_name=None, gdal_d
...
+        # check the file extension
+        assert gpkg_name[-5:] == '.gpkg', "The input file should end with .gpkg . Please check your input."
+
+        path, gpkg_name = parse_gpkg_path(path, gpkg_name)
+
+        return self.query_to_gpkg(f"select * from {schema}.{table}",
+                                 path=path, gpkg_name=gpkg_name, cmd=cmd, gdal_data_loc=gdal_data_loc, tbl_name = tbl_name,
+                                 print_cmd=print_cmd, srid=srid)
+    
+    def gpkg_to_table(self, path=None, table=None, schema=None, gpkg_name=None, cmd=None,
```
shostetter, line 2234: dont you need to specify the table being used in the geopackage?

pysqldb3/sql.py
shostetter, line 241: cant you reuse the SHP del index queries?

pysqldb3/tests/helpers.py
```diff
@@ -399,3 +399,27 @@ def set_up_xls():
...
+        'some_value': {0: 'test1', 1: 'test2'}
+    }
+    df = pd.DataFrame(data)
+    df.to_csv(os.path.join(DIR, "sample.csv"), index=False)
+    fle = os.path.join(DIR, "sample.csv")
+
+    cmd = f'''ogr2ogr -f "GPKG" {DIR}\\testgpkg.gpkg -dialect sqlite -sql 
+    "SELECT gid, GeomFromText(WKT, 4326) as geom, some_value FROM sample" {fle}'''
+    os.system(cmd.replace('\n', ' '))
+    print ('Sample geopackage ready...')
```
shostetter, line 416: i would add an assert to validate the gpkg file was created

pysqldb3/tests/helpers.py
```diff
@@ -399,3 +399,27 @@ def set_up_xls():
...
+
+def clean_up_geopackage():
+    fldr = os.path.join(os.path.dirname(os.path.abspath(__file__)))
+    for ext in ('gpkg'):
```
shostetter, line 420: the loop isnt needed if there is only 1 extension type

shostetter (commented):
I added a few comments throughout this, but one major point is i think the geopackage is more like a geodatabase than a shp since it can contain multiple tables, so you may want to review for that. Also, the tests arent passing because some of the setup files arent being created and there is some missing info.

cindyatnycdot:
I just pushed a lot of new edits to the Geopackage enhancement, such as bulk upload and overwriting features. All my tests had passed except for the one that uses a 'public' `default_schema` since I don't have write access to that schema, but I kept the test in so that it's consistent with the Shapefile tests.

pysqldb3/pysqldb3.py
cindyatnycdot: ok i will change this

pysqldb3/geopackage.py
```diff
@@ -75,8 +83,32 @@ def write_gpkg(self, tbl_name = None, print_cmd=False):
...
+        
+        elif not overwrite and os.path.isfile(os.path.join(self.path, self.gpkg_name)): # check if the geopackage already exists
+            
+            try:
+                exists_cmd = f'ogrinfo {os.path.join(self.path, self.gpkg_name)}'
+                ogr_response = subprocess.check_output(exists_cmd, stderr=subprocess.STDOUT)
+                table_exists = re.findall(f"{self.gpkg_tbl}", str(ogr_response)) # only allows tables names with underscores, numbers, and letters
+                assert len(table_exists) == 0, "The table name to be exported already exists in the geopackage. Either change to Overwrite = True or check the name of the table to be copied."
+
+            finally:
```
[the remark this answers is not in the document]
cindyatnycdot: The `try` statement is to see if my command returns a valid message, and if it does, it counts how many items are in the list.  And if it doesn't, then it creates an empty list and counts it as 0, rather than giving me an error. Maybe I don't need to use `finally` (?) but I didn't think it should be an `else:` because i want to come up with a number at the end.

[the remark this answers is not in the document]
cindyatnycdot: we do need to catch the error, because i can't count the number of entries in a list if the output isn't a list. as a specific example, i run a windows command that tells me what tables are in the geopackage. but if the geopackage isn't working or there's an error, i won't get a list of the tables and instead, i'll get an error message; in that case, i will just manually set the table list to an empty list. if the geopackage exists, i should return an output that i can parse for the table count. but if it doesn't exist, there is nothing to parse. that is why i have the try statement to catch the error.

pysqldb3/geopackage.py
```diff
@@ -171,38 +228,95 @@ def shp_to_gpkg(self, shp_name=None, gpkg_name=None, tbl_name=None, print_cmd=Fa
 
         return
 
-    def gpkg_to_shp(self, export_path = None, gpkg_name=None, print_cmd=False):
+    def gpkg_to_shp(self, print_cmd=False):
```
shostetter, line 231: this should take in an output path for the shape, it can default back to the same path as the geopackage, but that should be over writeable

pysqldb3/geopackage.py
```diff
@@ -10,14 +12,15 @@ class Geopackage:
     def __str__(self):
         pass
 
-    def __init__(self, dbo=None, path=None, table=None, schema=None, query=None, gpkg_name=None, cmd=None,
+    def __init__(self, dbo=None, path=None, table=None, schema=None, query=None, gpkg_name=None, gpkg_tbl = None, cmd=None,
```
shostetter, line 15: Most of the class variables seem too related to the methods and my or may not all be needed in all cases. the dbo param for example if you are doing geopackage to shp the db is not needed; and the table and geopackage parameters should be at the method level not the class, presumably the geopackage class is an instance for the entire geopackage which can include multiple tables and istn directly related to a database

pysqldb3/geopackage.py
```diff
@@ -75,8 +83,32 @@ def write_gpkg(self, tbl_name = None, print_cmd=False):
         if not self.path:
             self.path = file_loc('folder')
 
-        if not self.table:
-            self.table = tbl_name
+        if not self.gpkg_tbl:
+            self.gpkg_tbl = self.table
+
+        # overwrite vs update vs an issue has arisen
+        if overwrite:
```
shostetter, line 90: maybe check if it exists 1st then if not you can skip the rest and it will simplify the other conditions

pysqldb3/geopackage.py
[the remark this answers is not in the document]
shostetter: what happens if you dont have overwrite or update flags and you try to write to gpkg with a table that already exists? I assume it throws an error, why not return that error with a note about using the overwrite or update flags to the user - it should fail if you try it and there is a conflict and you havent explicitly stated overwrite or update

shostetter (commented):
left a structural comment that may change a lot of code, so pausing on the review for that

pysqldb3/pysqldb3.py
```diff
@@ -2103,9 +2108,9 @@ def query_to_gpkg(self, query, path=None, gpkg_name=None, cmd=None, tbl_name = N
         try:
             # Drop the temp table
             if self.type == PG:
-                self.query("drop table {}".format(tmp_table_name), internal=True, strict=False)
+                self.query(f"drop table {tmp_table_name}", internal=True, strict=False)
```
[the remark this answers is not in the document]
cindyatnycdot: i tried to replace the query with `drop_table` but `drop_table` doesn't work since the table in question is a temp table and the function expects a schema

pysqldb3/geopackage.py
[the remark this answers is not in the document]
shostetter: i dont see it in the exception

pysqldb3/geopackage.py
cindyatnycdot: ok i am thinking I will make them into `@staticmethods` and add extra arguments for input for things like `write_gpkg`, `read_gpkg`, and `query_to_gpkg`

cindyatnycdot:
Tests are passing for Hao now. Issues were that there was a geopackage file called `testgpkg.gpkg` in `pysqldb3/pysqldb3/tests` that needed to be deleted as well as temp tables under `ris/working/__temp_log_table_[your name]__` that should be cleared from PG (they hadn't been deleted because the tests had run into an error during an earlier run). Please check that any test geopackages are cleared from the `test` and `test_data` folders, and that any gpkg-related temp tables are also removed before running the tests. Thanks!

pysqldb3/geopackage.py
cindyatnycdot: If it exists, I don't want to automatically rewrite the file in case someone didn't realize they already had a geopackage with that name and had meant to rename their new file as something else.

pysqldb3/geopackage.py
```diff
@@ -201,69 +199,62 @@ def shp_to_gpkg(self, shp_name, gpkg_tbl = None, print_cmd=False):
         # check if the gpkg already exists. if so, we add another layer to the gpkg
         if not os.path.isfile(os.path.join(self.path, self.gpkg_name)):
```
shostetter, line 200: i think you should just create some methods that check if the geopackage exists and if a geopackage table exists, so you can just call that each time rather than reusing this code

pysqldb3/geopackage.py
```diff
@@ -201,69 +199,62 @@ def shp_to_gpkg(self, shp_name, gpkg_tbl = None, print_cmd=False):
...
         except subprocess.CalledProcessError as e:
             print("Ogr2ogr Output:\n", e.output)
             print('Ogr2ogr command failed. The Geopackage/feature class was not written.')
-            raise subprocess.CalledProcessError(cmd=print_cmd_string([self.dbo.password], self.cmd), returncode=1)
-
         return
 
-    def gpkg_to_shp(self, print_cmd=False):
-        
+    def gpkg_to_shp(self, gpkg_tbl = None, export_path = None, cmd = None):
```
shostetter, line 224: cmd param is not used, did you mean for it to be print_cmd?

pysqldb3/geopackage.py
```diff
@@ -201,69 +199,62 @@ def shp_to_gpkg(self, shp_name, gpkg_tbl = None, print_cmd=False):
...
         """
         Converts a Geopackage to a Shapefile.
         The output Shapefile name will match the name of the geopackage table to be copied.
 
-        :param gpkg_name: file name for geopackage input (should end in .gpkg)
-
-        :param export_path: The folder directory to place the shapefiles output.
+        :param gpkg_name (str): file name for geopackage input (should end in .gpkg)
+        :param gpkg_tbl (str): OPTIONAL specific table within the geopackage to convert to a Shapefile
```
shostetter, line 230: could this be a list if you only want a subset of tables?

pysqldb3/geopackage.py
```diff
@@ -201,69 +199,62 @@ def shp_to_gpkg(self, shp_name, gpkg_tbl = None, print_cmd=False):
...
+            export_path = self.path
 
         ### bulk upload capabilities ### 
         
-        if not self.gpkg_tbl:
+        if not gpkg_tbl:
 
             #######
             # if no table name listed in argument, then we assume bulk upload
             # this prompts us to count the number of tables within the geopackage to confirm bulk upload
```
shostetter, line 254: why do you need to count the tables?

pysqldb3/geopackage.py
```diff
@@ -273,11 +264,11 @@ def gpkg_to_shp(self, print_cmd=False):
             except subprocess.CalledProcessError as e:
```
shostetter, line 264: i assume this is due to different naming restrictions in geopackage vs shp, why not just sanitize the table names on output (or maybe gdal will do it for you?) otherwise wont this potentially skip some tables?  I think that could be problematic, so i would rather it fail if there is a name issue if renaming on export isnt an option, and maybe give you the table names that were the problem, then add a method to rename geopackage tables to this class so the user can fix it

pysqldb3/geopackage.py
```diff
@@ -290,79 +281,88 @@ def gpkg_to_shp(self, print_cmd=False):
 
             # if the bulk upload involves only 1 table and no db table name was specified, we name the table the same thing as the original gpkg table
             elif tbl_count == 1:            
-                gpkg_tbl_names = {re.search(r'([a-z0-9A-Z_]+)$', tables_in_gpkg[0]).group(): self.table}
+                gpkg_tbl_names = {re.search(r'([a-z0-9A-Z_]+)$', tables_in_gpkg[0]).group(): gpkg_tbl}
```
shostetter, line 284: FYI i believe `[\w]` is the same as  `[a-z0-9A-Z_]`, but simpler

pysqldb3/geopackage.py
```diff
@@ -290,79 +281,88 @@ def gpkg_to_shp(self, print_cmd=False):
...
                 print(ogr_response)
             except subprocess.CalledProcessError as e:
                 print("Ogr2ogr Output:\n", e.output)
                 print('Ogr2ogr command failed. The Geopackage/feature class was not written.')
-                raise subprocess.CalledProcessError(cmd=print_cmd_string([self.dbo.password], self.cmd), returncode=1)
         
         return
 
-    def table_exists(self):
+    def table_exists(self, dbo = None, schema = None, table = None):
```
shostetter, line 310: why is this wrapper needed, cant you just call the dbo.table exists?

pysqldb3/geopackage.py
```diff
@@ -290,79 +281,88 @@ def gpkg_to_shp(self, print_cmd=False):
...
+        :param table: table name
+        :param schema: schema name
         :return: boolean if gpkg name exists as a table in the schema
         """
 
-        return self.dbo.table_exists(table=self.table, schema=self.schema)
+        return dbo.table_exists(schema = schema, table = table)
 
-    def del_indexes(self):
+    def del_indexes(self, dbo, schema = None, table = None):
```
shostetter, line 321: why is this needed, if you drop the table the indexes are automatically dropped

pysqldb3/geopackage.py
```diff
@@ -290,79 +281,88 @@ def gpkg_to_shp(self, print_cmd=False):
...
+            indexes_to_delete = dbo.internal_data
 
             for _ in list(indexes_to_delete):
                 table_name, index_name, column_name, idx_typ = _
                 if 'pkey' not in index_name and 'PK' not in index_name:
-                    self.dbo.query('DROP INDEX {t}.{i}'.format(t=self.table, i=index_name), strict=False, internal=True)
+                    dbo.query('DROP INDEX {t}.{i}'.format(t=table, i=index_name), strict=False, internal=True)
 
-    def read_gpkg(self, precision=False, private=True, gpkg_encoding=None, print_cmd=False):
+    def read_gpkg(self, dbo = None, schema = None, table = None, port = 5432, srid = '2263', gdal_data_loc=GDAL_DATA_LOC, precision=False, private=True, gpkg_encoding=None, print_cmd=False):
```
shostetter, line 348: if the output is always a database table the dbo should be required. Also i think  private should default to False

pysqldb3/geopackage.py
```diff
@@ -374,15 +374,15 @@ def read_gpkg(self, precision=False, private=True, gpkg_encoding=None, print_cmd
...
-            self.del_indexes()
+            self.del_indexes(dbo = dbo, schema = schema, table = table)
 
-            print('Deleting existing table {s}.{t}'.format(s=self.schema, t=self.table))
-            self.dbo.drop_table(schema=self.schema, table=self.table)
+            print('Deleting existing table {s}.{t}'.format(s=schema, t=table))
+            dbo.drop_table(schema=schema, table=table)
 
         path = self.path
         full_path = os.path.join(path, self.gpkg_name)
```
shostetter, line 388: for organizational reasons i would say bulk read should be seperate functions that can get called by the read if needed

pysqldb3/geopackage.py
```diff
@@ -422,52 +422,52 @@ def read_gpkg(self, precision=False, private=True, gpkg_encoding=None, print_cmd
 
             # if the bulk upload involves only 1 table and no db table name was specified, we name the table the same thing as the original gpkg table
```
shostetter, line 423: seems like a lot of the same code from gpkg to shp, can you move to an internal function to DRY this up?

pysqldb3/geopackage.py
```diff
@@ -499,84 +499,87 @@ def read_gpkg(self, precision=False, private=True, gpkg_encoding=None, print_cmd
...
+                    dbo.query('grant select on {s}."{t}" to public;'.format(
+                        s=schema, t=output_name),
                         timeme=False, internal=True, strict=True)
                 except:
                     pass
-            self.rename_geom()
+            self.rename_geom(dbo, schema, table)
 
-    def rename_geom(self):
+    def rename_geom(self, dbo = None, schema = None, table = None, port=5432):
```
shostetter, line 531: can you just call the shapefile.rename_geom? or move that to util or something so they can both access it

pysqldb3/pysqldb3.py
```diff
@@ -1779,19 +1779,19 @@ def df_to_geojson(df):
         fig.show()
         return
 
-    def query_to_shp(self, query, path=None, shp_name=None, cmd=None, gdal_data_loc=GDAL_DATA_LOC,
-                     print_cmd=False, srid=2263, shp = True, gpkg_tbl = None):
+    def query_to_shp(self, query,  gpkg_tbl, path=None, shp_name=None, cmd=None, gdal_data_loc=GDAL_DATA_LOC,
```
shostetter, line 1782: 1st why is there here, i see a seperate function query_to_gpkg? 2nd if it is going to be in this then the gpkg_tbl needs to be an optional param

pysqldb3/pysqldb3.py
```diff
@@ -2085,25 +2085,26 @@ def feature_class_to_table(self, path, table, schema=None, shp_name=None, gdal_d
...
 
 
-    def query_to_gpkg(self, query, path=None, gpkg_name=None, cmd=None, gpkg_tbl = None, gdal_data_loc=GDAL_DATA_LOC,
+    def query_to_gpkg(self, query, gpkg_tbl = None, path=None, gpkg_name=None, cmd=None,  gdal_data_loc=GDAL_DATA_LOC,
                      print_cmd=False, srid=2263):
         """
         Exports query results to a geopackage (.gpkg) file.
         :param query: SQL query as string type
-        :param path: folder path for output gpkg
+        :param gpkg_tbl (str): Table name to be written in the Geopackage output
```
shostetter, line 2093: doesnt gpkg_name need to be non-optional for this?

pysqldb3/pysqldb3.py
```diff
@@ -2131,17 +2132,17 @@ def table_to_gpkg(self, table, schema=None, strict=True, path=None, gpkg_name=No
                                  path=path, gpkg_name=gpkg_name, cmd=cmd, gdal_data_loc=gdal_data_loc, gpkg_tbl = gpkg_tbl,
                                  print_cmd=print_cmd, srid=srid)
     
-    def gpkg_to_table(self, path=None, schema=None, table =None, gpkg_name=None, gpkg_tbl=None, cmd=None,
+    def gpkg_to_table(self, dbo = None, path=None, schema=None, table =None, gpkg_name=None, gpkg_tbl=None,
```
shostetter, line 2135: dbo and gpkg need to be required for this

shostetter (commented):
I havent looked at or run the tests yet, but i left a number of comments on the PR

pysqldb3/geopackage.py
```diff
@@ -45,23 +32,30 @@ def name_extension(name):
         else:
             return name + '.gpkg'
 
-    def write_gpkg(self, overwrite = False, print_cmd=False):
+    def write_gpkg(self, dbo = None, table = None, schema  = None, query = None, srid='2263', gdal_data_loc=GDAL_DATA_LOC, cmd = None, overwrite = False, print_cmd=False):
+        
         """
         Converts a SQL or Postgresql query to a new geopackage.
```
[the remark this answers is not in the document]
cindyatnycdot: It takes either. This function was modeled after the `write_shp` function, which is used for converting a SQL table to the output format, as well as a query to the output format. If you use `query_to_gpkg`, it actually calls `write_gpkg`. But as a standalone function, it makes it easy for users to convert a SQL table to geopackage.

pysqldb3/geopackage.py
```diff
@@ -45,23 +32,30 @@ def name_extension(name):
         else:
             return name + '.gpkg'
 
-    def write_gpkg(self, overwrite = False, print_cmd=False):
+    def write_gpkg(self, dbo = None, table = None, schema  = None, query = None, srid='2263', gdal_data_loc=GDAL_DATA_LOC, cmd = None, overwrite = False, print_cmd=False):
```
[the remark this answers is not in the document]
cindyatnycdot: Please see comment below

pysqldb3/geopackage.py
cindyatnycdot: Ah sorry, my previous comment is confusing.  `overwrite` refers to the table, but the command has to be very specific based on what you want to do, which is why the `if` statement has 3 options. Following from my previous comment, I didn't want to accidentally overwrite a table in an existing package if someone meant to rename it as something else. But maybe you don't think that is a common-enough issue to merit that functionality?

In your example where a gpkg exists but a certain table doesn't, you can't write the same command as if the geopackage doesn't exist. you have to explicitly include the term "-update" in order to append the table to the geopackage. Otherwise, it will overwrite your entire geopackage and include only the 1 table (which I could see as a huge headache if you spent time compiling/curating tables for the gpkg). The `if` statement currently checks for all those scenarios.

pysqldb3/pysqldb3.py
cindyatnycdot: There is a test called `test_query_to_gpkg_basic_pth` based on `test_query_to_shp_basic_pth` that tests for whether the function still works if a full path is given instead of a gpkg name.  So I allowed `gpkg_name` not to be written so that it would pass this test. Maybe we want to remove this functionality?

pysqldb3/geopackage.py
cindyatnycdot: I can't call the same `rename_geom` function because the `Shapefile` one only takes `self` as an argument while my gpkg version has `dbo`, `schema` and `table` arguments, since I moved these inputs out of the class. Would require me to add more arguments to the Shapefile.py file, which I can do if you want

pysqldb3/geopackage.py
cindyatnycdot: Will delete this. Had the count in an earlier iteration of the `.py` file

pysqldb3/geopackage.py
cindyatnycdot: I have changed the logic so that it cleans the name if possible, or returns an error that the user must clean the gpkg table names before trying to load the package

pysqldb3/geopackage.py
cindyatnycdot: (writing this comment for reference) I think what we agreed on was that if they weren't uploading all the tables in a geopackage, they would write a loop themselves to load the individual tables

pysqldb3/geopackage.py
```diff
@@ -245,119 +318,100 @@ def gpkg_to_shp(self, gpkg_tbl = None, export_path = None, cmd = None):
...
-                if 'pkey' not in index_name and 'PK' not in index_name:
-                    dbo.query('DROP INDEX {t}.{i}'.format(t=table, i=index_name), strict=False, internal=True)
+        return
 
-    def read_gpkg(self, dbo = None, schema = None, table = None, port = 5432, srid = '2263', gdal_data_loc=GDAL_DATA_LOC, precision=False, private=True, gpkg_encoding=None, print_cmd=False):
+    def read_gpkg(self, dbo, table = None, gpkg_tbl = None, schema = None, port = 5432, srid = '2263', gdal_data_loc=GDAL_DATA_LOC,
+                    precision=False, private=False, gpkg_encoding=None, print_cmd=False, bulk_upload = False):
         """
-        Reads a geopackage into SQL or Postgresql as a table
+        Reads a single geopackage table into SQL or Postgresql as a table
```
shostetter, line 402: this reads a table **or all of the tables** when bulk is true

pysqldb3/geopackage.py
```diff
@@ -374,90 +428,42 @@ def read_gpkg(self, dbo = None, schema = None, table = None, port = 5432, srid =
...
+            if not gpkg_tbl:
+                gpkg_tbl = self.gpkg_tbl
 
+            if table:
+                assert table == re.sub(r'[^A-Za-z0-9_]+', r'', table) # make sure the name will load into the database
+                table = table.lower()
+            else:
+                # clean the geopackage table name
+                table = gpkg_tbl.replace('.gpkg', '').lower()
+                table = re.sub(r'[^A-Za-z0-9_]+', r'', table) # clean the table name in case there are special characters
```
shostetter, line 445: you could wrap the table names in `"` so the special characters would be allowed, other wise if you have a table named `tbl_1` and `tbl_1*` you will get a collision in the database

pysqldb3/geopackage.py
```diff
@@ -521,13 +527,154 @@ def read_gpkg(self, dbo = None, schema = None, table = None, port = 5432, srid =
...
                 try:
-                    dbo.query('grant select on {s}."{t}" to public;'.format(
-                        s=schema, t=output_name),
+                    dbo.query(f'grant select on {schema}."{table}" to public;',
                         timeme=False, internal=True, strict=True)
                 except:
                     pass
             self.rename_geom(dbo, schema, table)
 
+        else:
```
shostetter, line 536: this is a lot of code, i would propose moving to a bulk function that is called here

pysqldb3/geopackage.py
```diff
@@ -521,13 +527,154 @@ def read_gpkg(self, dbo = None, schema = None, table = None, port = 5432, srid =
...
+
+            # loop through the tables     
+            for output_name, input_name in gpkg_tbl_names.items():
+            
+                if dbo.table_exists(table = output_name, schema = schema):
+
+                    print(f'Deleting existing table {schema}.{output_name}')
+                    dbo.drop_table(schema=schema, table=output_name)
+
+                if dbo.type == 'PG':
```
shostetter, line 571: this logic was used earlier, can you pull it out into a function that gets the list of table names so the logic only exists once

pysqldb3/geopackage.py
shostetter: ok I added this as an issue to fix https://github.com/safety-analytics-mapping/pysqldb3/issues/89

pysqldb3/pysqldb3.py
```diff
@@ -2172,45 +2178,126 @@ def gpkg_to_table(self, dbo = None, path=None, schema=None, table =None, gpkg_na
...
+                            gdal_data_loc=gdal_data_loc, precision = precision, private = private, print_cmd = print_cmd)
+            
+            for table in output_table_names.values():
 
-        gpkg = Geopackage(path=path, gpkg_name=gpkg_name, gpkg_tbl = gpkg_tbl)
+                if self.type == "MS":
 
-        gpkg.read_gpkg(dbo = dbo, schema = schema, table = table, gpkg_encoding = gpkg_encoding, srid = srid,
-                       gdal_data_loc=gdal_data_loc, precision = precision, private = private, print_cmd = print_cmd)
+            # rename geom columns if necessary (problem only identified in MS)
```
shostetter, line 2195: isnt there a rename geom function in the geopackage class that can be used here

pysqldb3/pysqldb3.py
```diff
@@ -2172,45 +2178,126 @@ def gpkg_to_table(self, dbo = None, path=None, schema=None, table =None, gpkg_na
...
-                if geom_output != 'geom':
-                    self.query(f"alter table {schema}.{table} rename column {geom_output} to geom")
-            except:
-                pass
-                    
-        self.tables_created.append(f"{schema}.{table}")
+            if self.type == "MS":
+            # rename geom columns if necessary (problem only identified in MS)
+                try:
+                    geom_output = self.dfquery(f""" SELECT COLUMN_NAME
```
shostetter, line 2235: same comment as above to use the rename geom function

pysqldb3/pysqldb3.py
```diff
@@ -2128,21 +2129,24 @@ def table_to_gpkg(self, table, schema=None, path=None, gpkg_name=None, gpkg_tbl
...
 
+        if not gpkg_tbl:
+            gpkg_tbl = table
+
         return self.query_to_gpkg(f"select * from {schema}.{table}",
                                  path=path, gpkg_name=gpkg_name, cmd=cmd, gdal_data_loc=gdal_data_loc, gpkg_tbl = gpkg_tbl,
                                  print_cmd=print_cmd, srid=srid)
     
-    def gpkg_to_table(self, dbo = None, path=None, schema=None, table =None, gpkg_name=None, gpkg_tbl=None, 
+    def gpkg_to_table(self, dbo, gpkg_name, gpkg_tbl, path=None, schema=None, table =None,
```
shostetter, line 2139: dbo isnt needed to be passed in since its in the self instance

pysqldb3/pysqldb3.py
```diff
@@ -2172,45 +2178,126 @@ def gpkg_to_table(self, dbo = None, path=None, schema=None, table =None, gpkg_na
...
+                        self.query(f"alter table {schema}.{table} rename column {geom_output} to geom")
+                except:
+                    pass
+                        
+            self.tables_created.append(f"{schema}.{table}")
+
+            if temp:
+                self.__run_table_logging([schema + "." + table], days=days)
+
+    def gpkg_to_table_bulk_upload(self, dbo, gpkg_name, path=None, schema=None,
```
shostetter, line 2265: dbo isnt needed as its in the self instance

pysqldb3/query.py
```diff
@@ -537,15 +537,15 @@ def print_query(query_string):
             print(_)
 
     @staticmethod
-    def query_to_gpkg(dbo, query, path=None, gpkg_name=None, cmd=None, gpkg_tbl=None, gdal_data_loc=GDAL_DATA_LOC, print_cmd=False, srid=2263):
+    def query_to_gpkg(dbo, query, path, gpkg_name, gpkg_tbl, cmd=None, gdal_data_loc=GDAL_DATA_LOC, print_cmd=False, srid=2263):
```
shostetter, line 540: why is this static?

pysqldb3/query.py
```diff
@@ -557,7 +557,7 @@ def query_to_gpkg(dbo, query, path=None, gpkg_name=None, cmd=None, gpkg_tbl=None
                         gpkg_tbl = gpkg_tbl)
         
         if not gpkg_name:
```
shostetter, line 559: why not set as optional with default as ''?

pysqldb3/pysqldb3.py
```diff
@@ -2172,45 +2178,126 @@ def gpkg_to_table(self, dbo = None, path=None, schema=None, table =None, gpkg_na
...
+                    try:
+                        geom_output = self.dfquery(f"""SELECT COLUMN_NAME 
+                                        FROM information_schema.COLUMNS 
+                                        WHERE data_type = 'USER-DEFINED'
+                                        AND lower(TABLE_NAME)=lower('{table}')
+                                        AND table_schema = '{schema}'
+                                """).values[0][0]
+                    
+                        if geom_output != 'geom':
+                            self.query(f"alter table {schema}.{table} rename column {geom_output} to geom")
```
shostetter, line 2218: all of these internally used query calls should include the `internal=True` flag  so they arent stored with the user generated queries, thats something that needs to be cleaned up generally in the code since that is a new flag

shostetter (commented):
this is looking good i added a number of comments and questions - i havent reviewed the tests yet, will save that for the end. 

One thing to check for is are you being consistent with the use of gpkg vs geopackage in function and variable names

pysqldb3/query.py
cindyatnycdot: Because `query_to_shp` was static so I thought it would remain consistent

pysqldb3/geopackage.py
```diff
@@ -245,119 +318,100 @@ def gpkg_to_shp(self, gpkg_tbl = None, export_path = None, cmd = None):
...
+
             except subprocess.CalledProcessError as e:
                 print("Ogr2ogr Output:\n", e.output)
                 print('Ogr2ogr command failed. The Geopackage/feature class was not written.')
         
         return
-
-    def table_exists(self, dbo = None, schema = None, table = None):
+    
+    def gpkg_to_shp_bulk_upload(self,
```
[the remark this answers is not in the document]
cindyatnycdot: OK I will rename

pysqldb3/geopackage.py
```diff
@@ -31,24 +30,91 @@ def name_extension(name):
...
+        Checks if a geopackage table already exists, in case it isn't meant to be overwritten.
+        """
+
+        try:
+            exists_cmd = f'ogrinfo {os.path.join(self.path, self.gpkg_name)}'
+            ogr_response = subprocess.check_output(exists_cmd, stderr=subprocess.STDOUT)
+            table_exists = re.findall(f"{gpkg_tbl}", str(ogr_response)) # only allows tables names with underscores, numbers, and letters
+        
+        except:
+            pass
```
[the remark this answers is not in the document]
shostetter: OK if its an internal function you should make it private (`__geopackage_tbl_exists` or semi-private `_geopackage_tbl_exists`), but what is the use case where it fails the try and you want it to continue?

pysqldb3/query.py
cindyatnycdot: I tried this but it doesn't work. I have tests (`test_query_to_gpkg_basic_pth`) that don't pass if I default `gpkg_name = ''`. Even though I don't fill in the argument in the test, my tests fail due to `gpkg_name` defaulting to 'None'.

pysqldb3/query.py
shostetter: odd i cant really imagine why it should be since it requires the dbo

pysqldb3/geopackage.py
```diff
@@ -261,21 +265,35 @@ def shp_to_gpkg(self, shp_name, gpkg_tbl = None, print_cmd = False):
             gpkg_tbl = shp_name.replace('.shp', '')
 
         # check if the gpkg already exists. if so, we add another layer to the gpkg
-        if not self.geopackage_exists(self.gpkg_name):
+        if not self.gpkg_exists(self.gpkg_name):
```
shostetter, line 268: rather than repeating the cmd code, why not use the if then to set the format variable (which i think are the only things that are changing) and then just call it once - meaning just use the if/then to set the _overwrite and _update variables

pysqldb3/geopackage.py
```diff
@@ -411,7 +423,7 @@ def read_gpkg(self, dbo, table = None, gpkg_tbl = None, schema = None, port = 54
         :param private:
         :param gpkg_encoding: encoding of data within Geopackage
         :param print_cmd: Optional flag to print the GDAL command that is being used; defaults to False
-        :param bulk_upload: Default to False. Use read_gpkg_bulk_upload() to upload all gpkg tables to db.
+        :param bulk_upload: Default to False. Do not set to True; please use read_gpkg_bulk() to upload all gpkg tables to db.
```
shostetter, line 426: doesnt look like this is actually a param that can be set, which makes sense given the comment

pysqldb3/pysqldb3.py
```diff
@@ -2136,19 +2136,17 @@ def table_to_gpkg(self, table, gpkg_name, gpkg_tbl = None, schema=None, path=Non
...
                      srid=2263, port=None, gdal_data_loc=GDAL_DATA_LOC, precision=False, private=False, temp=True,
                      gpkg_encoding=None, print_cmd=False, days=7, bulk_upload = False):
         """
         Imports single geopackage table to database. This uses GDAL to generate the table.
-        :param dbo: Database connection
-        :param path: File path of the shapefile
-        :param schema: Schema to use in the database (defaults to db's default schema)
         :param gpkg_name: Geopackage name (ends in .gpkg)
         :param gpkg_tbl: Input table name from Geopackage.
+        :param path: File path of the shapefile
```
shostetter, line 2146: should be path to geopackage, unless thats in the gpkg_name in which case this can be removed

shostetter (commented):
left a few questions

pysqldb3/geopackage.py
cindyatnycdot: I will just put all of it into a bulk upload function

pysqldb3/geopackage.py
```diff
@@ -31,24 +30,91 @@ def name_extension(name):
...
+    def geopackage_tbl_exists(self, gpkg_tbl):
+               
+        """
+        Checks if a geopackage table already exists, in case it isn't meant to be overwritten.
+        """
+
+        try:
+            exists_cmd = f'ogrinfo {os.path.join(self.path, self.gpkg_name)}'
+            ogr_response = subprocess.check_output(exists_cmd, stderr=subprocess.STDOUT)
+            table_exists = re.findall(f"{gpkg_tbl}", str(ogr_response)) # only allows tables names with underscores, numbers, and letters
```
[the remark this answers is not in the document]
cindyatnycdot: Yes, right now it will give false positives. I will fix the regex.

pysqldb3/geopackage.py
shostetter: oh ok i get it i do think true false is more intuitive

pysqldb3/geopackage.py
shostetter: it doesnt hurt cause it just a doc string, but whats the point of saying not to do something that you cant do anyway

pysqldb3/geopackage.py
cindyatnycdot: I have cleaned it up in my next update

pysqldb3/geopackage.py
```diff
@@ -37,20 +37,24 @@ def list_gpkg_tables(self):
...
+                return tables_in_gpkg
+                
+            except subprocess.CalledProcessError as e:
+                print(e)
+        
+        else:
+             print("This geopackage file does not exist at this file location.")
     
-    def geopackage_exists(self, gpkg_name = None):
+    def gpkg_exists(self, gpkg_name = None):
```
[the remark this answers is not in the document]
cindyatnycdot: It defaults in the next line of the code where i have `if not gpkg_name: gpkg_name = self.gpkg_name`. I cannot set `gpkg_name  = self.gpkg_name` as the function's default parameter

pysqldb3/geopackage.py
```diff
@@ -63,28 +67,27 @@ def geopackage_exists(self, gpkg_name = None):
...
-            exists_cmd = f'ogrinfo {os.path.join(self.path, self.gpkg_name)}'
-            ogr_response = subprocess.check_output(exists_cmd, stderr=subprocess.STDOUT)
-            table_exists = re.findall(f"{gpkg_tbl}", str(ogr_response)) # only allows tables names with underscores, numbers, and letters
         
-        except:
-            pass 
-        
-        if len(table_exists) == 0:
-            gpkg_tbl_exists = False
+        if self.gpkg_exists():
```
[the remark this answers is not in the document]
cindyatnycdot: I only ever run this function when I've already confirmed that the geopackage exists, but I wanted to build in an extra check just in case. I will just set it to False if no gpkg_exists

cindyatnycdot:
@shostetter I've merged this branch with `dev`, removed the `rename_geom` function from `geopackage.py` and run all the gpkg-related tests to make sure it's working. Ready for review now.

cindyatnycdot:
Noticed in the commits that I somehow deleted Test.xlsx but I can't seem to add it back as a commit to my PR....

pysqldb3/geopackage.py
```diff
@@ -264,36 +268,28 @@ def shp_to_gpkg(self, shp_name, gpkg_tbl = None, overwrite = False, print_cmd =
...
-                                                full_path = self.path,
-                                                shp_name = shp_name,
-                                                _overwrite = '-overwrite',
-                                                gpkg_tbl = gpkg_tbl)
-
+            _overwrite = '-overwrite'
         elif not self.gpkg_tbl_exists(gpkg_tbl): # if gpkg exists and overwrite was not explicitly called
+            _update = '-update' # then add the table to the gpkg
+        elif self.gpkg_tbl_exists(gpkg_tbl): # if the gpkg and table exists but no overwrite was called
+            print("The table name to be copied to the geopackage already exists. Either change to Overwrite = True or check the name of the table to be copied.")
```
shostetter, line 283: I think this if elif conditional funnel will work, but i think explicitly checking the logic for each section will be better for long term maintenance, less likely future changes could break it. 
so you might consider the simpler approach of something like this 
```
if not overwrite and not self.gpkg_tbl_exists(gpkg_tbl:
then ...

if not overwrite and self.gpkg_tbl_exists(gpkg_tbl:
then... 
```

pysqldb3/geopackage.py
```diff
@@ -343,41 +339,24 @@ def gpkg_to_shp(self,
                 count_cmd = COUNT_GPKG_LAYERS.format(full_path = os.path.join(self.path, self.gpkg_name))
                 ogr_response = subprocess.check_output(shlex.split(count_cmd.replace('\n', ' ')), stderr=subprocess.STDOUT)
                 tables_in_gpkg = re.findall(r"\\n\d+:\s(.*?)(?=\\r|\s\(.*\))", str(ogr_response)) # only allows tables names with underscores, numbers, and letters
```
shostetter, line 341: i think rather than have the bulk call this function with bulk = true, i would move the bulk logic to the gpkg_to_shp_bulk function and have this call that

pysqldb3/geopackage.py
```diff
@@ -456,26 +437,55 @@ def read_gpkg(self, dbo, table = None, gpkg_tbl = None, schema = None, port = 54
...
             table = table.lower()
 
-            if dbo.table_exists(schema = schema, table = table):
+            gpkg_tbl_names[table] = gpkg_tbl
 
-                print(f'Deleting existing table {schema}.{table}')
-                dbo.drop_table(schema=schema, table=table)
+        else:
+
+            ####### BULK UPLOAD ########
```
shostetter, line 444: It seems to me this is doing too many things, it should only import 1 table, if there are many tables that need importing it should go to a bulk read function, or if you want you can set the input table to be a list and loop through that. 
i think rather than have the bulk call this function with bulk = true, i would move the bulk logic to the bulk read function and have this call that

pysqldb3/pysqldb3.py
```diff
@@ -2355,3 +2355,108 @@ def gpkg_to_table_bulk(self, gpkg_name,  path=None, schema=None,
                             days=days,
                             bulk_upload = True)
 
+    def backup_table(self, org_schema, org_table, backup_path, backup_schema, backup_table):
```
shostetter, line 2358: odd why is this showing up in the dif?

shostetter (commented):
I left some comments, mostly organizational

pysqldb3/geopackage.py
cindyatnycdot: OK I have changed it in my latest update

pysqldb3/geopackage.py
cindyatnycdot: OK I have changed that in my latest update

shostetter (approved):
looks good to me!

cindyatnycdot:
Merging. I realized that the "Test.xlsx" file that I deleted from `dev` is actually not supposed to be in `dev` anyway, and is a file generated from `test_xls_to_table` that wasn't deleted properly

PR #71, opened 2024-09-16
Add schema_tables function

colinandersonnyc:
creates function that gets list of all tables in specified schema


colinandersonnyc:
Addresses https://github.com/safety-analytics-mapping/pysqldb3/issues/70

PR #75, opened 2024-09-30
73 table backup to txt file function

shostetter:
Adds function to create backup sql file for table 
Adds function to create table from backup sql file
Adds function to get indexes associated with table

### DIFF
--- a/pysqldb3/tests/test_backup_table.py
+++ b/pysqldb3/tests/test_backup_table.py
@@ -74,6 +74,7 @@ class TestBackupTablesPg:
         # run backup
         db.drop_table(pg_schema, test_pg_from_backup)
         schema_table_name = db.create_table_from_backup(test_back_file)
+        assert re.findall(r'[\-["\w"\]]*\.[-\["\w"\]]*', schema_table_name)
 
         # validate table exists
         assert db.table_exists(test_pg_from_backup, schema=pg_schema)


cindyatnycdot:
Hi Seth, I ran into errors whenever I tried to input PG (PG only, MS was fine) tables with capitals like `Test_pg_to_backup_` or 'Test_pg_from_backup` instead of `test_pg_to_backup`, so some more standardizing of cases due to user input is needed.

I also noticed that if i have an existing table with the same name as the backup for upload, it will overwrite the table without warning that a table with that same name exists. maybe that's how it's intended to work. i thought i'd point that out in case people want to be warned that they're about to replace an existing table

shostetter:
@cindyatnycdot you should add tests that check for these issue, they should fail and i can update the code to make them pass

cindyatnycdot:
I've added new tests to check for this issue. The Postgres one fails while the MS SQL one does not; it might be related to how the programs format table names.

pysqldb3/tests/test_backup_table.py
```diff
@@ -0,0 +1,523 @@
...
+        if os.path.isfile(test_back_file):
+            os.remove(test_back_file)
+        assert not os.path.isfile(test_back_file)
+
+        db.backup_table(pg_schema, test_pg_to_backup, test_back_file, pg_schema, test_pg_from_backup)
+        assert os.path.isfile(test_back_file)
+
+        # run backup
+        db.drop_table(pg_schema, test_pg_from_backup)
+        schema_table_name = db.create_table_from_backup(test_back_file)
```
cindyatnycdot, line 76: I fixed this in my branch for the capitalized table names (so no need to fix it in this version, but I'm letting you know), but I wanted to point out that the `schema_table_name` variable is never referenced again and doesn't appear in any other test, so I removed the variable creation.

shostetter: added a check that its returning correctly

cindyatnycdot (approved):

PR #79, opened 2024-10-07
Create tests for tables with capitalized names

cindyatnycdot:
Redoing the PR

### DIFF
--- a/pysqldb3/tests/test_backup_table.py
+++ b/pysqldb3/tests/test_backup_table.py
@@ -73,8 +73,8 @@ class TestBackupTablesPg:
 
         # run backup
         db.drop_table(pg_schema, test_pg_from_backup)
-        schema_table_name = db.create_table_from_backup(test_back_file)
-        assert re.findall(r'[\-["\w"\]]*\.[-\["\w"\]]*', schema_table_name)
+
+        db.create_table_from_backup(test_back_file)
 
         # validate table exists
         assert db.table_exists(test_pg_from_backup, schema=pg_schema)
@@ -281,11 +281,11 @@ class TestBackupTablesPg:
         assert not os.path.isfile(test_back_file)
 
     def test_backup_tables_capitalized(self):
-        db.drop_table(table=test_pg_to_backup.capitalize(), schema=pg_schema)
+        db.drop_table(table=test_pg_to_backup, schema=pg_schema)
 
         # table schema
         db.query(f"""
-            CREATE TABLE {pg_schema}.{test_pg_to_backup.capitalize()} (
+            CREATE TABLE {pg_schema}.{test_pg_to_backup} (
                 id int,
                 column2 text,
                 column3 timestamp,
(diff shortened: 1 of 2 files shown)


pysqldb3/tests/test_backup_table.py
```diff
@@ -278,6 +278,60 @@ def test_messy_table(self):
         if os.path.isfile(test_back_file):
             os.remove(test_back_file)
         assert not os.path.isfile(test_back_file)
+
+    def test_backup_tables_capitalized(self):
+        db.drop_table(table=test_pg_to_backup.capitalize(), schema=pg_schema)
+
+        # table schema
+        db.query(f"""
+            CREATE TABLE {pg_schema}.{test_pg_to_backup.capitalize()} (
```
shostetter, line 287: i dont think this will work, if the table name isnt wrapped in `""` or [] for MS the table name will not be capitalized in the database

pysqldb3/tests/test_backup_table.py
```diff
@@ -278,6 +278,60 @@ def test_messy_table(self):
...
+        """)
+        # populate table
+        for i in range(10):
+            db.query(f"""
+                INSERT INTO {pg_schema}.{test_pg_to_backup.capitalize()}
+                    (id, column2, column3, "1 test messy column")
+                values ({i}, '{i}', '{'2022-10-14 10:12:40-04'}', '{'test '*i}')
+            """)
+        # validate table created
+        assert db.table_exists(test_pg_to_backup.capitalize(), schema=pg_schema)
```
shostetter, line 302: this will also currently fail (which is fine for now) since table exists doesnt account for capitalized names, you should probably just recreate the underlying sql logic to check if table exists and account for capitalized name

shostetter (commented):
left some comments

pysqldb3/tests/test_backup_table.py
cindyatnycdot: when i made these tests, i wasn't that concerned about the actual letter case for the tables that get read into the database (like whether the table in the db is actually called "table123" or "Table123"). rather, i flagged that if i were a user who used the `backup_table` function and typed "Table123" instead of "table123" as the table name argument, the `backup_table` function in PG wouldn't work and would error out. This only happens in the PG test; MS SQL seems to standardize the name whether you use capitals or not.

maybe this is more of a question of how you want the function to work. Should the backup tables maintain the lettercase for the input (in turn, i should add the "" quotes in), or should the function just standardize the input table name (make everything lowercase regardless of how you enter it)? 

I can easily change the tests to remove the `.capitalize()` suffix from all the entries except the function call (if we should just standardize the name to lowercase). let me know what you think!

shostetter: but if the table in the db is actually named table123 than backing up Table123 should error since that table doesn't exist, just like if i type table321 instead of table123. 

the backup function should preserve the table name it is in the database, so if the name is capitalized in the database it should be in the back, but if not then it should match the formatting of the table. 

right now almost everything in pysqldb3 will standardize to lowercase by default, there are some edge cases that will keep the input format but they are rare. I do want to go though and fix it so everything is lower unless specified that case is important but that hasnt been done yet

cindyatnycdot: OK i've altered the PG `test_backup_tables_capitalized` test to change the way the table name is capitalized and the test passes now. Yes, not the biggest issue but I was surprised when the same test passed for one server type but not the other due to capitalizing the table name.

pysqldb3/tests/test_backup_table.py
cindyatnycdot: I know you made #79 to address the capitalization in `table_exists` and it looks like it works. I had copied over the code manually into my local version, tested the backup tables with the revised `table_exists` logic, and then made sure not to upload the `table_exists` code change since it'll change once #79 gets merged.

cindyatnycdot:
@shostetter  Hi Seth, I looked into why the MS tests aren't passing and it has to do with the `assert dbo.table_exist()' command within the `create_table_from_backup` function (because that's where it fails). I had a hard time diagnosing the exact issue or how to fix it, but I suspect it has to do with the way the table argument input is formatted, because the table exists but it can't find it.

image

shostetter:
> @shostetter Hi Seth, I looked into why the MS tests aren't passing and it has to do with the `assert dbo.table_exist()' command within the `create_table_from_backup` function (because that's where it fails). I had a hard time diagnosing the exact issue or how to fix it, but I suspect it has to do with the way the table argument input is formatted, because the table exists but it can't find it.
> 
> image

odd, what happens if you run table exists on a table you know is there?

cindyatnycdot:
> > @shostetter Hi Seth, I looked into why the MS tests aren't passing and it has to do with the `assert dbo.table_exist()' command within the `create_table_from_backup` function (because that's where it fails). I had a hard time diagnosing the exact issue or how to fix it, but I suspect it has to do with the way the table argument input is formatted, because the table exists but it can't find it.
> > image
> 
> odd, what happens if you run table exists on a table you know is there?

The table name doesn't automatically get square brackets around it when the table name is fed into `table_exists` so I added some lines into `pysqldb3.py` that got the tests passing. Not sure if this would cause other tests to fail.

shostetter:
@cindyatnycdot can we close this now?

PR #86, opened 2024-10-31
Sample tests for sql_to_sql functions

cindyatnycdot:
- This branch is built off `dev` and therefore doesn't have the `sql_to_sql` functions in the branch yet, as you had requested.
- Several of the tests were failing when I had tried the tests in your `84-data-io-needs-sql-to-sql` branch. The issues were that that geometry columns get renamed when they are copied from one server/db to another (not sure if you want to change the way the function works so that it doesn't change the column name from `geom` to `ogr_geometry`, datetime column in one of the test tables seemed to have an issue, and there were dtypes that seemed to have gotten changed when copying the table.

### DIFF
--- a/pysqldb3/data_io.py
+++ b/pysqldb3/data_io.py
@@ -433,7 +433,7 @@ def sql_to_sql(from_sql, to_sql, org_table, LDAP_from=False, LDAP_to=False, spat
     # enforce column data types in dest match the source
     for c in cols.keys():
         if '-' in c:
-            # GDAL sanitizes `'` out of column names this will put them back so the start/end tables match
+            # GDAL sanitizes `-` out of column names this will put them back so the start/end tables match
             to_sql.query(f"""EXEC sp_rename '{dest_schema}.{dest_table}.{c.replace('-', '_')}', '{c}', 'COLUMN';""")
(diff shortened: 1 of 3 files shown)


shostetter:
Thanks @cindyatnycdot i made some code changes for these tests, can you please review?

pysqldb3/data_io.py
```diff
@@ -421,9 +421,31 @@ def sql_to_sql(from_sql, to_sql, org_table, LDAP_from=False, LDAP_to=False, spat
...
+
+    cols = from_sql.get_table_columns(org_table, schema=org_schema)
+    cols = {c[0]:c[1] for c in cols}
+    cols_to = to_sql.get_table_columns(dest_table, schema=dest_schema)
+    cols_to = {c[0]: c[1] for c in cols_to}
+
+    # enforce column data types in dest match the source
+    for c in cols.keys():
+        if '-' in c:
+            # GDAL sanitizes `'` out of column names this will put them back so the start/end tables match
```
cindyatnycdot, line 436: for this comment, do you mean that "# GDAL sanitizes '-' out of column names " instead?

shostetter:
@cindyatnycdot if this looks good please merge

PR #95, opened 2024-11-19
Citext backup table test

cindyatnycdot:
Added a test. I noticed that all the MS tests were also failing for me now due to the "text"-type columns; not sure if this is also failing for you.

### DIFF
--- a/pysqldb3/pysqldb3.py
+++ b/pysqldb3/pysqldb3.py
@@ -741,10 +741,22 @@ class DbConnect:
             columns = '*'
         else:
             if self.type == PG:
-                columns = "column_name, data_type"
+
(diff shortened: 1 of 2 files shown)


cindyatnycdot:
My MS tests are passing now but the one PG test with citext type columns isn't. The error I get is that it doesn't recognize 'citext' columns (even though I can create citext column tables in ris3 in PgAdmin).  I saw online that I should run `CREATE EXTENSION IF NOT EXISTS citext WITH SCHEMA working;` before running the query, but I don't have user authorization to do so.

shostetter:
@cindyatnycdot  this is good to merge now, right?

PR #99, opened 2024-11-27
fix for qry creates table into - with () and + in query logic

shostetter:


Horaceshi624 (approved):
looks good, gained more flexibility. Just one question: after remove '?' at line 348, If a query "select into" a public table without specifying the schema, will this function be able to catch the schema?

shostetter:
> looks good, gained more flexibility. Just one question: after remove '?' at line 348, If a query "select into" a public table without specifying the schema, will this function be able to catch the schema?

ah it was redundant, the sds variable is {0,3} so it must be there between 0 and 3 times so the ? wasnt needed since it includes 0 in the range

PR #106, opened 2024-12-11
Add pg_to_sql_qry function

cindyatnycdot:


pysqldb3/data_io.py
```diff
@@ -94,6 +94,94 @@ def pg_to_sql(pg, ms, org_table, LDAP=False, spatial=True, org_schema=None, dest
...
+
+    try:
+        ogr_response = subprocess.check_output(shlex.split(cmd.replace('\n', ' ')), stderr=subprocess.STDOUT)
+        print(ogr_response)
+    except subprocess.CalledProcessError as e:
+        print("Ogr2ogr Output:\n", e.output)
+        print('Ogr2ogr command failed.')
+        raise subprocess.CalledProcessError(cmd=print_cmd_string([ms.password, pg.password], cmd), returncode=1)
+
+    ms.tables_created.append(dest_schema + "." + dest_table)
```
shostetter, line 180: the new format for this should be `ms.tables_created.append((ms.server, ms.database, dest_schema, dest_table))`

pysqldb3/tests/test_io.py
```diff
@@ -221,44 +221,210 @@
...
-#
-#
+
+
+class TestPgtoSqlQry:
+
+    @classmethod
+    def setup_class(cls):
+        helpers.set_up_test_table_pg(db)
+
```
shostetter, line 265: doesnt see like you are using this

pysqldb3/tests/test_io.py
```diff
@@ -221,44 +221,210 @@
...
+        db.query(f"""
+                    create table {pg_schema}.{test_pg_to_sql_qry_table} (test_col1 int, test_col2 int);
+                    insert into {pg_schema}.{test_pg_to_sql_qry_table} VALUES(1, 2);
+                    insert into {pg_schema}.{test_pg_to_sql_qry_table} VALUES(3, 4);
+        """)
+
+        # pg_to_sql_qry
+        data_io.pg_to_sql_qry(db, sql, query=
+                             f"""
+                             select * from {pg_schema}.{test_pg_to_sql_qry_table}
```
shostetter, line 283: maybe make the select something more specific than select * with no filter to really ensure its tested correctly

pysqldb3/tests/test_io.py
```diff
@@ -221,44 +221,210 @@
...
+        """)
+
+        data_io.pg_to_sql_qry(db, sql, query=f"""
+                                               SELECT * --comments within the query
+                                                FROM {pg_schema}.{test_pg_to_sql_qry_spatial_table} -- geom here
+                                                -- end here""",
+                              dest_table= test_pg_to_sql_qry_spatial_table,
+                              dest_schema = sql_schema)
+
+        data_io.pg_to_sql_qry(db, sql, query=f"""-- comments within the query
```
shostetter, line 390: why is this here, i thought it was a test of the spatial data transfer?

shostetter (commented):
looks good i left a few minor comments

cindyatnycdot:
@shostetter  I revised the spatial table tests (for `pg_to_sql_qry` and `sql_to_pg_query`) to remove the non-spatial tables that were also in the test. The test called `test_sql_to_pg_qry_spatial` doesn't work like I want it to (the test passes right now, but only because the input column is not a geometry column). 

The problem is that I can't copy a spatial table from sql_to_pg without there being an issue with the way that the pg table's geom column turns out. Doesn't matter whether i set `spatial = True` in the function call. There's no error when it runs, but the table does not have a working geom (it will say binary or null under the geometry column). I wonder if this was a known issue because the original test for `sql_to_pg_qry` doesn't compare the geometries between the pg and sql tables, and there is no test written for a spatial table that's directly moved from sql to pg.

Wanted to get your thoughts

PR #113, opened 2025-01-09
Allow PG tables with arrays to be backed up

cindyatnycdot:
I could only get the test to work if the arrays had double quotes around the text like '{"value1", "value2"}'


shostetter (commented):
looks good, i would add a test that the array works as expected and the values are the same, i would suggest using unnest

PR #123, opened 2025-01-22
update dataframe_to_table_schema

colinandersonnyc:
override now checks against cleaned column name, not pre cleaned column name

### DIFF
--- a/pysqldb3/pysqldb3.py
+++ b/pysqldb3/pysqldb3.py
@@ -971,10 +971,13 @@ class DbConnect:
             if 'date' in col_name.lower() and col_type in ('int', 'bigint', 'float'):
                 col_type = 'varchar(500)'
 
+            #clean col_name to check against overrides and insert
+            col_name=clean_column(col_name)
+
             if column_type_overrides and col_name in column_type_overrides.keys():
-                input_schema.append([clean_column(col_name), column_type_overrides[col_name]])
+                input_schema.append([col_name, column_type_overrides[col_name]])
             else:
-                input_schema.append([clean_column(col_name), col_type])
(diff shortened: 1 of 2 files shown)


colinandersonnyc:
https://github.com/safety-analytics-mapping/pysqldb3/issues/122

pysqldb3/pysqldb3.py
```diff
@@ -971,10 +971,13 @@ def dataframe_to_table_schema(self, df, table, schema=None, overwrite=False, tem
             if 'date' in col_name.lower() and col_type in ('int', 'bigint', 'float'):
                 col_type = 'varchar(500)'
 
+            #clean col_name to check against overrides and insert
+            col_name=clean_column(col_name)
```
shostetter, line 975: Thanks this looks like it should work.
is there anywhere else the col_name is used that might expect the original value? Have you run the tests to validate there are no unintended effects? 

also the PRs should go to dev

shostetter (commented):
left a question

pysqldb3/pysqldb3.py
colinandersonnyc: Updated to go to dev. I'll look deeper into the col_name issue. How do I run the validation tests?

shostetter: you should run all the files in the tests directory with pytest or at least the ones you know will touch this function

colinandersonnyc:
fixed relevant tests. Also found a further error for the MS version of test_bulk_csv_to_table_basic_blank_col. It was trying to use test7.csv which isn't created anymore?

shostetter:
@colinandersonnyc  all tests are passing, so i think this is good to go

PR #136, opened 2025-02-11
Fix regex for "with cte"

cindyatnycdot:


cindyatnycdot:
Rebased the branch and confirmed that tests are still passing

shostetter (approved):
looks good, thanks!

PR #150, opened 2025-04-01
Reimagine geospatial functions/class to be simpler

### LINKED ISSUE #81: Revise methods for gpkg, shp, and feature class
Rethink the Geospatial functions and combine the various files

### LINKED ISSUE #142: Create function called "geospatial_read"
Function should create a db table from some sort of geospatial file (geodb, geopackage, shp)

### LINKED ISSUE #143: Create function called "geospatial_write"
Function that writes a db query to a geospatial format of the user's choice

### LINKED ISSUE #144: Create a function called "geospatial_convert"
Function that converts a geospatial file to another file type (such as gpkg -> shp or shp -> gpkg)

### LINKED ISSUE #145: Create a function that lists geospatial tables within a geodatabase
Lists all the tables within a geospatial database like gpkg
maybe it also lists whether the tables are empty (which is possible)

cindyatnycdot:


cindyatnycdot:
As a heads up, I had to turn `__run_table_logging` into `run_table_logging` in the `util.py` in order to move all the functions into `geospatial.py` (rather than have duplicate functions in `pysqldb3.py`).

Also, I added one extra commit after i requested review (to test that users could supply only a file path rather than a file path + input_file name to run the functions), so you should make sure to pull the latest commit from this branch. thanks!

shostetter (commented):
I added some comments on initial thoughts - happy to discuss if its helpful

pysqldb3/geospatial.py
```diff
@@ -0,0 +1,850 @@
...
+    :param days: if temp=True, the number of days that the temp table will be kept. Defaults to 7.
+    :param print_cmd: Optional flag to print the GDAL command that is being used; defaults to False
+    :return:
+    """
+
+    if input_file:
+        assert input_file.endswith('.shp') or input_file.endswith('.gpkg') or input_file.endswith('.dbf'), "The input file should end with .gpkg, .shp, or .dbf"
+        assert path, "Fill in the file path to the input file"
+    else:
+        assert path.endswith('.shp') or path.endswith('.gpkg') or path.endswith('.dbf'), "The path should end with .gpkg, .shp, or .dbf"
```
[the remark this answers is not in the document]
cindyatnycdot: When I was consolidating the original functions into the new functions, I thought that including `.dbf` as an import format would take care of the old `feature_class_to_table` function. I incorporated `.dbf` as a file format across multiple functions for consistency. 

Granted, I've never used the `.dbf` file format and wasn't sure how widely used it is.  Let me know if you have thoughts.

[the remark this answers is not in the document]
cindyatnycdot: I think I meant for it to say `.gdb`  . I had included this as part of folding the feature class function into the centralized function and got the extension wrong

pysqldb3/geospatial.py
```diff
@@ -0,0 +1,850 @@
...
+                                    col=col_name[2:-2], short_col=shortened_col)
+                elif dbo.type == MS:
+                    shortened_col = col_name[1:-1][:7]
+                    return_cols += " , cast([{col}] as date) [{short_col}_dt], cast(cast([{col}] as time) as varchar)" \
+                                   " [{short_col}_tm] ".format(
+                                    col=col_name[1:-1], short_col=shortened_col)
+
+        # Wrap the original query and select the non-datetime/timestamp columns and the parsed out dates/times
+        qry = f"select {return_cols} from ({query}) q "
+
```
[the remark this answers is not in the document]
cindyatnycdot: yes you're right, i only need 2 conditions

pysqldb3/geospatial.py
```diff
@@ -0,0 +1,850 @@
...
+    if output_file.endswith('.gpkg') and "." in output_file[:-5]:
+        output_file = output_file[:-5].replace(".", "_") + ".gpkg"
+        print(' The "." character is not allowed in output gpkg file names. Any "." have been removed.')
+    elif not output_file.endswith(".gpkg") and "." in output_file[-5]:
+        output_file = output_file[:-5].replace(".", "_") + '.gpkg'
+        print(' The "." character is not allowed in output gpkg file names. Any "." have been removed.')
```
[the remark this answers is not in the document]
cindyatnycdot: yes i am going to delete the if statement

pysqldb3/geospatial.py
```diff
@@ -0,0 +1,850 @@
...
+        _update = ''
+        _overwrite = ''
+    
+    elif not overwrite and gpkg_exists(path = path, file_name = output_file): # check if the geopackage already exists
+        
+        table_exists = gpkg_tbl_exists(path = path, gpkg_name = output_file, gpkg_tbl = gpkg_tbl)
+        
+        if table_exists == True:
+            print("The table name to be exported already exists in the geopackage. Either change to Overwrite = True or check the name of the table to be copied.")
+            exit
```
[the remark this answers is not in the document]
cindyatnycdot: like .gdb?

pysqldb3/geospatial.py
```diff
@@ -0,0 +1,850 @@
...
+            pass
+
+    rename_geom(dbo, schema, table)
+    dbo.tables_created.append((dbo.server, dbo.database, schema,  table))
+    
+    if temp:
+        dbo.run_table_logging([schema + "." + table], days=days)
+
+
+def input_gpkg_bulk(path, dbo, input_file = None, schema = None, port = 5432, srid = '2263', gdal_data_loc=GDAL_DATA_LOC,
```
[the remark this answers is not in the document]
cindyatnycdot: yes i am going to add that in

cindyatnycdot:
I rebased the branch after I finalized all my edits, so that I could incorporate any additional changes from DEV into my branch, and there weren't any conflicts that needed to be resolved (just some extra code that had been written unrelated to my work, and some blank lines deleted/added).

shostetter (commented):
After a super cursory review i have some high level comments/questions: 
1. Can you change run_table_logging to _run_table_logging so its semi private instead of making it a public method? 
2. i think there are some things that could be generalized more
> a. gpkg_exists could check for geopackage, shp, or geodatabase
> b. gpkg_tbl_exists could be generalized to work for geopackage and geodatabase
> b. can input_geospatial_bulk be folded into geospatial convert and can it include geodatabases? 
3. in convert do we need the input/output path to be different from the file? cant it be blahblah.gdb/table or blahblahblah.shp ? 
> a. why does gpkg_tbl need to be a different param from input_file? how are geodatabase tables handled? same thing for input_geospatial_file
4. is there a way to have a simpler routing function that is basically geo_something (niput data, output data, misc other needed params) that would figure out if its moving to a db or out of a db or converting etc? 
5. can we keep the query_to_shp etc from pysqldb and just phave them route to the geospatial functions - this will make it more backwards compatable

cindyatnycdot:
@shostetter - Replies to your feedback

1. I changed it to semi-private and my tests are passing so I can make that change.

2a. Yes, I can make that change so that `gpkg_exists` is applicable to .shp and .gdb files

2b. I didn't make `gpkg_tbl_exists` available to gdb because we cannot write things to a geodatabase (we discussed how we never could get FileGDB  installed correctly), so there is no point checking for tables within `.gdb` files to overwrite since I can't write to them anyway.

2c. What is the motivation behind folding `input_geospatial_bulk` into `geospatial_convert`? The former is for bulk uploading tables into a PG or MS database, while the latter is for converting to another geospatial format and doesn't require a database connection.

3. Maybe there is some confusion about how the function works. The output path is only necessary if you want to save the file in a different folder location; the argument  is optional. If you want to save it as the same file name in the same folder, you DO need to write the file name out under the argument `output_file`. Are you thinking that people would fill it in as `geospatial_convert(input_file = 'C:/Documents/cindy.gpkg', '.shp')`? I was envisioning that people would do `geospatial_convert(input_file = 'C/Documents/cindy.gpkg', 'cindy.shp')`

3a. The `gpkg_tbl` name can be different from the geopackage name and having it separate meant that people could convert only certain gpkg tables to a shapefile instead of all of them. Under `cindy.gpkg`, I could have 10 tables and one of them could be `lion` and that's the table I want to convert. For geodatabases, there is a `feature_class` argument where the user specifies the .shp file they want to input. I didn't put them under the same argument because I thought that might be confusing for the user; hard to come up with a single term that would apply to both circumstances.

4. It can be done... I don't see any major problems with doing that

5. sure, i can add those function names back in. other than `query_to_shp`, are the other ones you want to keep:
   - `query_to_gpkg`
   - `feature_class_to_table`
   - `shp_to_table`
   - `table_to_shp`

shostetter:
> @shostetter - Replies to your feedback
> 
> 1. I changed it to semi-private and my tests are passing so I can make that change.
> 
> 2a. Yes, I can make that change so that `gpkg_exists` is applicable to .shp and .gdb files
> 
> 2b. I didn't make `gpkg_tbl_exists` available to gdb because we cannot write things to a geodatabase (we discussed how we never could get FileGDB installed correctly), so there is no point checking for tables within `.gdb` files to overwrite since I can't write to them anyway.
> 
> 2c. What is the motivation behind folding `input_geospatial_bulk` into `geospatial_convert`? The former is for bulk uploading tables into a PG or MS database, while the latter is for converting to another geospatial format and doesn't require a database connection.
> 
> 3. Maybe there is some confusion about how the function works. The output path is only necessary if you want to save the file in a different folder location; the argument  is optional. If you want to save it as the same file name in the same folder, you DO need to write the file name out under the argument `output_file`. Are you thinking that people would fill it in as `geospatial_convert(input_file = 'C:/Documents/cindy.gpkg', '.shp')`? I was envisioning that people would do `geospatial_convert(input_file = 'C/Documents/cindy.gpkg', 'cindy.shp')`
> 
> 3a. The `gpkg_tbl` name can be different from the geopackage name and having it separate meant that people could convert only certain gpkg tables to a shapefile instead of all of them. Under `cindy.gpkg`, I could have 10 tables and one of them could be `lion` and that's the table I want to convert. For geodatabases, there is a `feature_class` argument where the user specifies the .shp file they want to input. I didn't put them under the same argument because I thought that might be confusing for the user; hard to come up with a single term that would apply to both circumstances.
> 
> 4. It can be done... I don't see any major problems with doing that
> 5. sure, i can add those function names back in. other than `query_to_shp`, are the other ones you want to keep:
>    
>    * `query_to_gpkg`
>    * `feature_class_to_table`
>    * `shp_to_table`
>    * `table_to_shp`

@cindyatnycdot  responding to some of your comments 

2b. it could still be useful to quickly check if something exists, which could be used when trying to import something or just check if it exists to validate expectations 

2c. sorry i think i meant to use `input_geospatial_file` im thinking if you dont provide a table table and you are importing from gdb or gpkg it would bulk import all of it

3. i think i was trying to reduce the number of parameters, so if the input info and export info always has both file path and file name there are fewer parameters to deal with 
3a. i was thinking of including the geopackage table or feature class table as part of the input/output path (ex. `geospatial_convert(input_file = 'C:/Documents/cindy.gpkg/table_name' ) then you dont need extra parameters for it 
5. yes that looks like a good list

pysqldb3/geospatial.py
```diff
@@ -114,6 +115,7 @@ def write_geospatial(dbo, path, output_file = None, table = None, schema = None,
     # assert that a valid file format was input
     if output_file:
         assert output_file.endswith(('.gpkg', '.shp', 'dbf')), "Output file needs to be .gpkg, .dbf, or .shp format"
```
shostetter: why dbf?

pysqldb3/pysqldb3.py
```diff
@@ -1900,368 +1899,108 @@ def df_to_geojson(df):
         fig.show()
         return
 
-    def query_to_shp(self, query,  gpkg_tbl = None, path=None, shp_name=None, cmd=None, gdal_data_loc=GDAL_DATA_LOC,
-                     print_cmd=False, srid=2263, shp = True):
+    def query_to_shp(self, query, shp_name = None, path=None, cmd=None, gdal_data_loc=GDAL_DATA_LOC,
```
shostetter, line 1912: shouldn't you extend the kwargs to allow for different export types? puls by removing or changing the params its no longer backwards compatible. I think ideally you would be able to pass any call that used to work to the new one and it would just route it  to the new geospatial function like you are doing

pysqldb3/pysqldb3.py
```diff
@@ -1900,368 +1899,108 @@ def df_to_geojson(df):
...
-        if temp:
-            self.__run_table_logging([schema + "." + table], days=days)
-
-    def feature_class_to_table(self, path, table = None, schema=None, shp_name=None, gdal_data_loc=GDAL_DATA_LOC,
-                               srid=2263, private=False, temp=True, fc_encoding=None, print_cmd=False,
-                               days=7, skip_failures=''):
+        write_geospatial(dbo = self, path = path, query = query, output_file = shp_name,
+                            cmd = cmd, gdal_data_loc = gdal_data_loc, print_cmd = print_cmd, srid = srid)
+        
+    def feature_class_to_table(self, path, table, shp_name = None, feature_class = None, schema=None,
```
shostetter, line 1929: same comment as query_to_shp

pysqldb3/tests/test_feature_class_to_table.py
```diff
@@ -1,357 +0,0 @@
...
-                        user=config.get('SQL_DB', 'DB_USER'),
-                        password=config.get('SQL_DB', 'DB_PASSWORD'),
-                        allow_temp_tables=True)
-
-fgdb = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'test_data/lion/lion.gdb')
-fc = 'node'
-table = 'test_feature_class_{}'.format(db.user)
-
-
-class TestFeatureClassToTablePg:
```
shostetter, line 31: shouldnt these tests still work?

pysqldb3/util.py
```diff
@@ -374,46 +374,26 @@ def print_cmd_string(password_list, cmd_string):
...
-
-    if path.endswith('.shp'):
-        shp = os.path.basename(path)
-        path = path.replace(shp, '')
+        return path, file_name
 
-    if not shp_name:
-        shp_name = shp
+    geospatial_exts = ('.shp', '.dbf', '.gpkg', '.tar', '.gz', '.tgz', '.tar.gz', '.7z', '.rar', '.gdb', '.zip')
+    if path.lower().endswith(geospatial_exts):
```
shostetter: not sure this will work. For geodatabases and i think geopackages you get something like `blahblahblah.gdb/feature_class_table_name` so it doesnt end with  gdb and if its zippped then you get even further `blah.zip/blahblahblah.gdb/feature_class_table_name`

shostetter (commented):
I haven't dug into the tests, but i think all of the old tests should still run since you saved the function and just pass them to the new geospatial, its probably worth having those tests there. I left some other small comments as well

pysqldb3/geospatial.py
cindyatnycdot: There is a test in the original `test_query_to_shp.py` file where the output is exclusively `.dbf` so i thought i would formally include it as an output format in the function

pysqldb3/tests/test_feature_class_to_table.py
cindyatnycdot: I moved these tests to `test_geospatial.py` so that there wouldn't be so many test files all related to geospatial functions. if you search for `TestFeatureClassToTablePg` in the file, you will find them

pysqldb3/pysqldb3.py
cindyatnycdot: I will mark this as resolved because we addressed in Microsoft Teams that the kwargs match the original function's and I didn't have additional/updated ones to incorporate into this function

pysqldb3/pysqldb3.py
cindyatnycdot: It has the same arguments as the original function. I added `extra_cmd` as one additional argument since the underlying function `input_geospatial_file` includes that.

pysqldb3/util.py
cindyatnycdot: If there is a feature class and they use `feature_class_to_table`, the user is expected to input it as an argument. This is similar to the current `feature_class_to_table` function where the user must input the feature class under the `shp_name` argument. That means that the input path should never end on a non-extension.

But for file paths where there can be 2 extensions (.zip and .gdb for example), I've changed the code to accommodate reading those files in.

cindyatnycdot:
Hi Seth, I addressed your outstanding comments and made some changes to the code to accommodate file paths like `filename.zip/smallerfolder/anotherdb.gdb`. I also rebased my branch since the PR said that there were conflicts between my branch and new updates to DEV, which added a bunch of commits. Maybe when this is finalized I will create a new branch so the commit history is cleaner if you want.

pysqldb3/geospatial.py
shostetter: are you writing to dbf files? I dont know if that is ever used

cindyatnycdot:
Let me know if you're envisioning creating even more `util` functions than what I've created. I have also changed `input_geospatial_file` to be called `upload_geospatial`... let me know what you think. I've also removed all `.dbf` related work from the functions and removed their 2 related tests

PR #151, opened 2025-04-08
140 add constraints to backup table

shostetter:
Backup table function did not include constraints, this fixes that - one problem is if you backup the reference table and change the name it will break this since the constraint will try to reference to the old table name. I dont think there is a way to mange this since the FK reference table wont be backed up in the same call


pysqldb3/tests/test_backup_table.py
```diff
@@ -672,3 +750,97 @@ def test_messy_table(self):
...
+        sql.query(f"select count(*) cnt from {ms_schema}.{test_sql_from_backup}")
+        assert sql.data[0][0] == 10
+
+        # Validate schema matches
+        _to = sql.get_table_columns(test_sql_to_backup, schema=ms_schema)
+        _from = sql.get_table_columns(test_sql_from_backup, schema=ms_schema)
+        assert _to == _from
+
+        # validate constraints exist
+        constraints_from = sql._get_table_constraints(ms_schema, test_sql_from_backup)
```
Horaceshi624, line 831: I might miss it but I didn't see where did you create test_sql_from_backup? (from my test run to_table successfully returned constraint info but from not)

so I also wrote my own code to test my new _get_table_constraint() and it seems work. maybe can you test run the 'test_backup_table.py' on my branch (140-add-constraints-to-backup-table-hao))?

pysqldb3/tests/test_backup_table.py
```diff
@@ -430,6 +430,84 @@ def test_table_with_array_contents(self):
...
+
+        if os.path.isfile(test_back_file):
+            os.remove(test_back_file)
+        assert not os.path.isfile(test_back_file)
+
+        db.backup_table(pg_schema, test_pg_to_backup, test_back_file, pg_schema, test_pg_from_backup)
+        assert os.path.isfile(test_back_file)
+
+        # run backup
+        db.drop_table(pg_schema, test_pg_from_backup)
```
Horaceshi624, line 484: I might miss it but I didn't see where did you create test_pg_from_backup?

pysqldb3/tests/test_backup_table.py
```diff
@@ -672,3 +750,97 @@ def test_messy_table(self):
...
+        # validate table created
+        assert sql.table_exists(test_sql_to_backup, schema=ms_schema)
+        sql.query(f"select count(*) cnt from {ms_schema}.{test_sql_to_backup}")
+        assert sql.data[0][0] == 10
+
+        if os.path.isfile(test_back_file):
+            os.remove(test_back_file)
+        assert not os.path.isfile(test_back_file)
+
+        sql.backup_table(ms_schema, test_sql_to_backup, test_back_file, ms_schema, test_sql_from_backup)
```
Horaceshi624, line 812: Hi Seth,
I tested this function on my end but it couldn't make backup table correctly (I guess i might have some issue with 'test_back_file'). Instead I compared your _get_table_constraint and mine and they are generating the same output now. maybe could you test run the code on this branch(where I updated _get_table_constraint): 140-add-constraints-to-backup-table-hao

shostetter: can you provide additional details on what was wrong with the backup table that is getting created?

Horaceshi624: The contents of the table look fine, but when I run your get_table_constraint() function, it throws an error saying the constraints for from_table are NoneType, meaning they don't exist. I also tested my modified version of the function and it returns an empty list. It's possible that the constraints weren’t copied over during the backup.

Horaceshi624 (approved):

PR #164, opened 2025-05-02
139 csv xls to table   enhancement

shostetter:
Adds process to remove empty columns with no headers from csv and xls to table, including ones with only ' ' in the data


cindyatnycdot (approved):
Looks great

PR #166, opened 2025-05-07
Add process to read compressed file

Horaceshi624:
Hi!
I added a process to detect if path or input_file include suffix in '.tar', '.gz', '.tgz', '.7z', '.rar' and if so we can use python packages like py7zr to decompress them automaticly and import shp inside. (note: shp name should be the same as compressed file name)

let me know if there's any questions


pysqldb3/geospatial.py
```diff
@@ -503,9 +510,13 @@ def gpkg_to_shp_bulk(   input_path,
...
 
-def input_geospatial_file(dbo, path, input_file = None, schema = None, table = None, feature_class = False, gpkg_tbl = None, port = 5432,
-                            srid = '2263', gdal_data_loc=GDAL_DATA_LOC, precision=False, private=False, encoding=None, zip = False, skip_failures = '',
-                            temp = True, days = 7, print_cmd=False):
+
+
+def input_geospatial_file(dbo, path, input_file=None, schema=None, table=None, feature_class=False, gpkg_tbl=None,
+                          port=5432,
+                          srid='2263', gdal_data_loc=GDAL_DATA_LOC, precision=False, private=False, encoding=None,
+                          zip=False, skip_failures='',
```
cindyatnycdot: When Seth was reviewing my Geospatial code, he said that maybe we could just get rid of the `zip`  argument and just detect the file/folder type via the `input_file` or `path`.

pysqldb3/geospatial.py
```diff
@@ -529,13 +540,55 @@ def input_geospatial_file(dbo, path, input_file = None, schema = None, table = N
...
+            raise FileNotFoundError("No .shp file found after decompression.")
+
+        shp_path = shp_files[0]
+
+        # Return the folder path, .shp filename, and temp dir for later cleanup
+        path, input_file, temp_dir = str(shp_path.parent), shp_path.name, temp_dir
+
+
+
+
```
cindyatnycdot: i think this needs to go after the `if input_file` section so that it checks the `input_file` field first. you might need to change the `assert` statements to check for your file extensions

pysqldb3/geospatial.py
```diff
@@ -529,13 +540,55 @@ def input_geospatial_file(dbo, path, input_file = None, schema = None, table = N
...
+                archive.extractall(path=temp_dir)
+        elif suffix == '.rar':
+            with rarfile.RarFile(full_path) as archive:
+                archive.extractall(path=temp_dir)
+        else:
+            shutil.rmtree(temp_dir)
+            raise ValueError(f"Unsupported compression format: {suffix}")
+
+        # Look for the first .shp file extracted
+        shp_files = list(Path(temp_dir).rglob('*.shp'))
```
cindyatnycdot: with these additional file types, is it possible that a folder would have more than 1 shapefile?

pysqldb3/tests/helpers.py
```diff
@@ -302,10 +304,21 @@ def set_up_shapefile():
                 z.write(filePath, os.path.basename(filePath))
     print('Sample zipped shapefile ready...')
 
+    component_paths = [
```
cindyatnycdot, line 308: can you add a comment above the section you added to describe what is happening? Similar to the comment in line 297 where it says `# Add shpfile to zip for testing`

cindyatnycdot (commented):
Hi Hao, here are my preliminary comments. I've been working on edits in my branch at the same time so we will later have to figure out what we want to do about rebasing your branch.

pysqldb3/geospatial.py
Horaceshi624: yeah I can add zip to my if/eilf condition and execute its importing code. Does it make sense?

pysqldb3/geospatial.py
Horaceshi624: my logic now can only get the shapefile having the same name as compressed folder

cindyatnycdot: is it possible for the shapefile not to match? or for more than 1 in a folder? we would want to be able to include all these scenarios. if there is more than 1, maybe we would want to add to the `input_gpkg_bulk` function

pysqldb3/geospatial.py
Horaceshi624: I see, I think it makes sense to decompress the file first then see if inside files are ended with legal suffix. If I checked suffix at the beginning then if the file inside compressed folder isn't shp gpkg or dbf I wouldn't have a way raise error. but I will double check to see if there's better way to do it

pysqldb3/geospatial.py
Horaceshi624: ok I updated it like this (to be in consistant as reading zip): this logic only applies when the path includes the name of the compressed file and the input file is the name of the specific SHP file we need to import.

pysqldb3/geospatial.py
cindyatnycdot: yes, you could use `example.endswith('zip')` or something like that to check if it's true

pysqldb3/geospatial.py
cindyatnycdot: ok yes that makes sense

pysqldb3/geospatial.py
Horaceshi624: yeah, I replaced the old if statement

pysqldb3/geospatial.py
```diff
@@ -2,6 +2,14 @@
 import subprocess
 import re
 import os
+import shutil
```
cindyatnycdot: you can delete this line. i accidentally added it when i was rebasing, and i realize now that it's already in line 8

pysqldb3/geospatial.py
```diff
@@ -539,7 +589,10 @@ def input_geospatial_file(dbo, path, input_file = None, schema = None, table = N
     else:
         precision = ''
 
-    if path.endswith('.zip') or input_file.endswith('.zip'):
+    print('check zip', os.path.splitext(path)[1].lower())
```
cindyatnycdot: should this be under line 594? because it seems like it would print "check zip" even if a path is intentionally not a zip. either it should be under the line 594 if statement or under a different if statement

pysqldb3/util.py
```diff
@@ -384,8 +384,9 @@ def parse_geospatial_file_path(path=None, file_name=None):
     # type: (str, str)
     if not path:
         return path, file_name
-    
-    if path.endswith('.shp') or path.endswith('.gdb') or path.endswith('.gpkg'):
+
+    geospatial_exts = ('.shp', '.dbf', '.gpkg', '.tar', '.gz', '.tgz', '.tar.gz', '.7z', '.rar')
```
cindyatnycdot: add .gdb

pysqldb3/tests/helpers.py
```diff
@@ -302,10 +304,22 @@ def set_up_shapefile():
...
+    with py7zr.SevenZipFile(os.path.join(DIR, 'test.7z'), 'w') as archive:
+        for filePath in component_paths:
+            archive.write(filePath, arcname=os.path.basename(filePath))
+    print('Sample 7z shapefile ready...')
+
 
 def clean_up_shapefile():
     fldr = os.path.join(os.path.dirname(os.path.abspath(__file__)))
-    for ext in ('shp', 'dbf', 'shx', 'prj', 'zip'):
+    for ext in ('shp', 'dbf', 'shx', 'prj', 'zip', '.tar', '.gz', '.tgz', '.7z', '.rar'):
```
cindyatnycdot: add `.gdb`

pysqldb3/util.py
cindyatnycdot: do we also need to add `.zip`?

pysqldb3/geospatial.py
Horaceshi624: oh I think I should delete it bc that was only for my tests before

pysqldb3/tests/test_geospatial.py
```diff
@@ -1392,6 +1392,47 @@ def test_read_shp_zip(self):
...
+    def test_read_shp_other_compressed(self):
+        fp = FOLDER_PATH + "\\test.7z"
+        shp_name = "test.shp"
+
+        # Make sure table doesn't already exist
+        db.drop_table(pg_schema, test_read_shp_table_name)
+        assert not db.table_exists(schema=pg_schema, table=test_read_shp_table_name)
+
+        # Read the .7z archive directly into PostGIS
+        s.input_geospatial_file(
```
cindyatnycdot, line 1404: i don't believe there's any code in the tests to remove the .7z file after you are done testing

cindyatnycdot (commented):
can you remove the zip=True from the `test_read_shp_zip` test? when you ran that test, was it passing? because it is failing for me due to that argument

pysqldb3/tests/test_geospatial.py
Horaceshi624: I think I added .7z and other types to clean_up_shapefile(), helper.py line 322
and sorry I made a copy of test_geospatial with pg part only and I might miss some update, let me double check

cindyatnycdot: you should still add a test for MS if it's not there, and i will test it on my end and let you know. since we would want the file to be compatible with both

cindyatnycdot: and yes i see that you added .7z to the cleanup. thanks for letting me know :)

cindyatnycdot:
@Horaceshi624  I figured out why the tests for Feature Class -> MS were going slow (it seems like that's just the way it is even from the PROD version), so the only thing you need to do is add the extra 7z test and we should be good to go!

cindyatnycdot (approved):

PR #173, opened 2025-09-02
170 Create table attribute consistency

shostetter:
SOme imports and bulk imports were not consistent in the format of the tables_created attribute, this standardizes across methods of creating tables and database types. All tables created should be listed as `(server, database, schema, table)`


shostetter:
TODO close this - replaced by #174

PR #175, opened 2025-10-02
Change get_table_columns and create_query

cindyatnycdot:


pysqldb3/pysqldb3.py
```diff
@@ -749,9 +750,41 @@ def get_schemas(self):
...
+    def get_table_columns(self, table_or_query, is_query = False, schema=None, full=False):
+        
         if not schema:
             schema = self.default_schema
+
+        if is_query:
+            # Makes a temp table name
+            tmp_table_name = f"tmp_query_to_geospatial_{self.user}_{str(datetime.datetime.now())[:16].replace('-', '_').replace(' ', '_').replace(':', '')}"
+
+            try:
```
shostetter: why not just use drop table if exists?

pysqldb3/pysqldb3.py
```diff
@@ -774,30 +807,123 @@ def get_table_columns(self, table, schema=None, full=False):
                           else DATA_TYPE end as DATA_TYPE
                       """
 
-        if self.type == PG:
+        # query columns
+        if self.type == PG and not is_query:
```
shostetter: why do you need not is_query? the whole point of creating a temp table is to be able to use this existing code...

pysqldb3/pysqldb3.py
```diff
@@ -774,30 +807,123 @@ def get_table_columns(self, table, schema=None, full=False):
...
+            ## CINDY: maybe combine with the previous if statement? need to change schema and table name
+            # turning it into query instead of dfquery also makes it harder because 
+            # we would need .data_columns and .data  and then we'd also need to map these separate outputs together
+            col_df = self.dfquery(f"""
+                SELECT *
+                FROM
+                INFORMATION_SCHEMA.COLUMNS
+                WHERE TABLE_NAME = '{tmp_table_name}'
+            """, internal = True)
+
```
shostetter: this should not be needed

pysqldb3/pysqldb3.py
```diff
@@ -774,30 +807,123 @@ def get_table_columns(self, table, schema=None, full=False):
...
+            
+            # querying the temp table is very different in MS
+            col_df = self.dfquery(f"""
+                SELECT
+                    [column] = c.name,
+                    [type] = t.name, 
+                    c.max_length, 
+                    c.precision, 
+                    c.scale, 
+                    c.is_nullable
```
shostetter: can you add something to the if for the column variabledefinition for sql and query to like the below to make it match the results for table and query? 

`c.name as column_name, 
case 
when t.name = 'text' then t.name
when t.max_length is not null 
then t.name + ' ('+ cast(t.max_length as varchar)+')' 
when t.max_length >= 3000 then t.name + ' (3000)' 
else t.name end as DATA_TYPE`

then reuse the columns variable and just change the from of the query

pysqldb3/pysqldb3.py
```diff
@@ -774,30 +807,123 @@ def get_table_columns(self, table, schema=None, full=False):
...
+                    t.system_type_id = t.user_type_id
+                WHERE
+                    [object_id] = OBJECT_ID(N'tempdb.dbo.#{tmp_table_name}');
+                """, internal=True)
+
+        if not is_query:
+            # this is the desired output for a table input
+            results = self.__get_most_recent_query_data(internal=True)
+
+        else:
```
shostetter: none of this belongs here, this function should only be doing 1 thing, take in a table or query and return a list of column names and data types. That is something that can be reused for many things. The date type checker is specific to exporting geo data so that logic belongs where you are doing that work, not here

shostetter (commented):
I left a few comments

pysqldb3/pysqldb3.py
```diff
@@ -749,14 +750,45 @@ def get_schemas(self):
...
+        
         if not schema:
             schema = self.default_schema
+
+        if is_query:
+            # check if allow_temp_tables is on if a query is being used. if not, turn it on
+            initial_temp_status = self.allow_temp_tables 
+            self.allow_temp_tables = True
+            # Makes a temp table name
+            tmp_table_name = f"tmp_query_to_geospatial_{self.user}_{str(datetime.datetime.now())[:16].replace('-', '_').replace(' ', '_').replace(':', '')}"
```
shostetter: not critical but the temp table name shouldnt have geospatial in it, this function is to be used beyond that process

pysqldb3/pysqldb3.py
```diff
@@ -749,14 +750,45 @@ def get_schemas(self):
...
+
+        # Drop the temp table if it already exists
+            if self.type == PG:
+                self.query(f"drop table if exists {tmp_table_name}", internal=True, strict=False)
+            elif self.type == MS:
+                self.query(f"drop table if exists #{tmp_table_name}", internal=True, strict=False)
+
```
shostetter: you can move this into the query or at least the if statement below to simplify

pysqldb3/pysqldb3.py
```diff
@@ -765,39 +797,79 @@ def get_table_columns(self, table, schema=None, full=False):
...
-        return self.__get_most_recent_query_data(internal=True)
+        elif self.type == MS and is_query:
+            
+            # querying the temp table is very different in MS
+            self.query(f"""
+                SELECT
+                    c.name as column_name, 
+                    case    when t.name in ('text', 'int', 'date') then t.name
+                            when t.max_length is not null and t.max_length < 3000 then t.name + ' ('+ cast(t.max_length as varchar)+')'  
+                            when t.max_length >= 3000 then t.name + ' (3000)'
```
shostetter, line 846: why do you need this carve out for 3000?

shostetter (commented):
left a few minor comments and questions

pysqldb3/pysqldb3.py
cindyatnycdot: that was how it was done in lines 800-804 (where it was capped at 3000 cars) so I copied it.

pysqldb3/util.py
```diff
@@ -824,108 +825,59 @@ def create_query(dbo, query):
     """
     Create and clean the query as part of write_geospatial
```
shostetter, line 824: if this function is just for the geospatial it shouldn't be in util, also it should be named to clarify that it is creating a query for geo output

pysqldb3/util.py
```diff
@@ -824,108 +825,59 @@ def create_query(dbo, query):
...
     """
+   
+    # find datetime fields
+    col_df = dbo.get_table_columns(query, is_query = True)
```
shostetter, line 828: this isnt actually a df, right?

pysqldb3/util.py
```diff
@@ -824,108 +825,59 @@ def create_query(dbo, query):
...
-            col_df[col_df['type'].str.contains('datetime') | col_df['type'].str.contains('timestamp')]['column'])]
+        left_bracket = '['
+        right_bracket = ']' 
+
+    cols = [left_bracket + c[0] + right_bracket for c in col_df]
+    dt_col_names = [c[0] for c in col_df if (('datetime' in c[1]) | ('timestamp' in c[1]))]
 
     # Make string of columns to be returned by select statement
-    return_cols = ' , '.join([c for c in cols if c not in dt_col_names])
+    results = ' , '.join([c for c in cols if c not in dt_col_names])
```
shostetter, line 845: sorry to be pedantic, but this isn't really results, its a column name string or something

pysqldb3/util.py
```diff
@@ -824,108 +825,59 @@ def create_query(dbo, query):
...
 
     # If there are datetime/timestamp columns:
     if len(dt_col_names) > 0:
+
         if dbo.type == PG:
-            print_cols = str([str(c[2:-2]) for c in dt_col_names])
+            col_range = 2
+        elif dbo.type == MS:
+            col_range = 1
```
shostetter: can you explain this? why do you need 2 different ranges?

shostetter: what if there are multiple datetime columns?

pysqldb3/util.py
```diff
@@ -824,108 +825,59 @@ def create_query(dbo, query):
...
-            print_cols = str([str(c[1:-1]) for c in dt_col_names])
+        print_cols = str([str(c[col_range:-col_range]) for c in dt_col_names])
 
         print(f"""
         The following columns are of type datetime/timestamp: \n
         {print_cols}
         
-        Shapefiles don't support datetime/timestamps with both the date and time. Each column will be split up
+        Shapefiles and GPKG don't support datetime/timestamps with both the date and time. Each column will be split up
         into colname_dt (of type date) and colname_tm (of type **string/varchar**).
```
shostetter, line 857: why are converting to sting, shouldnt it be split into a date field and a time field?

shostetter (commented):
the get_table_columns changes look good, but the util changes shouldn't really be in this PR

PR #177, opened 2025-10-09
Update write_geospatial

cindyatnycdot:
Simplify function, update tests, and make sure everything is working

### DIFF
--- a/pysqldb3/geospatial.py
+++ b/pysqldb3/geospatial.py
@@ -7,21 +7,20 @@ from .cmds import *
 from .sql import *
 from .util import *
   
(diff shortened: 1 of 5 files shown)


shostetter:
@cindyatnycdot shouldnt there be more deleted rows from utils? I think `create_query` is still there

shostetter (commented):
why dont i see anything deleted from the util file?

cindyatnycdot:
> why dont i see anything deleted from the util file?

Just deleted `create_query` now. I thought maybe we wanted to delete things from the `util.py` file at the very end.

shostetter (approved):

PR #179, opened 2025-11-12
Geospatial new tests dtypes

cindyatnycdot:
Following our last pair programming session, my to-dos were to:
1. query the user's data types for geometry (I included integer as well) instead of hardcoding them since they can vary by user and we want the tests to work for all users.
2. Add tests for the datetime fields so we know that they are working in PG and MS
3. Edit `write_geo_cmd_query` so that we didn't need an extra line of code to remove columns where the `data_type is None`. This was happening because MS temp tables don't recognize Geometry fields and would set them to None.

### DIFF
--- a/pysqldb3/geospatial.py
+++ b/pysqldb3/geospatial.py
@@ -109,8 +109,7 @@ def write_geo_cmd_query(dbo, query_or_table, is_query = False, schema = ''):
     columns_with_brackets = [left_bracket + c[0] + right_bracket for c in columns_with_types]
 
     # identify date columns if they are in the intended output
-    dt_col_names = [c for c in columns_with_types if c[1] is not None] ## TODO - what is this doing/is it needed?
-    dt_col_names = [c[0] for c in dt_col_names if (('datetime' in c[1]) | ('timestamp' in c[1]))] # after, check if there is datetime / timestamp
+    dt_col_names = [c[0] for c in columns_with_types if (('datetime' in c[1]) | ('timestamp' in c[1]))] # after, check if there is datetime / timestamp
 
     # if there are no date columns, we can query immediately
     if len(dt_col_names) == 0:
--- a/pysqldb3/geospatial.py
+++ b/pysqldb3/geospatial.py
@@ -131,7 +131,7 @@ def write_geo_cmd_query(dbo, query_or_table, is_query = False, schema = ''):
                                 col=col_name, shortened_col = shortened_col
                                 )
             elif dbo.type == MS:
-                results += " , convert(datetime, [{col}]) [{shortened_col}_dt], convert(varchar, convert(time, [{col}]))" \
+                results += " , convert(date, [{col}]) [{shortened_col}_dt], convert(varchar, convert(time, [{col}]))" \
                                 " [{shortened_col}_tm] ".format(
                                 col=col_name, shortened_col = shortened_col
                                 )
--- a/pysqldb3/geospatial.py
+++ b/pysqldb3/geospatial.py
@@ -131,7 +131,7 @@ def write_geo_cmd_query(dbo, query_or_table, is_query = False, schema = ''):
                                 col=col_name, shortened_col = shortened_col
                                 )
             elif dbo.type == MS:
-                results += " , cast([{col}] as date) [{shortened_col}_dt], cast(cast([{col}] as time) as varchar)" \
+                results += " , convert(datetime, [{col}]) [{shortened_col}_dt], convert(varchar, convert(time, [{col}]))" \
                                 " [{shortened_col}_tm] ".format(
                                 col=col_name, shortened_col = shortened_col
                                 )
@@ -221,6 +221,11 @@ def write_geospatial(dbo, path,  table = None, schema = '', query = None, gpkg_t
     
     # update allows you to add an extra table into an existing geopackage
         _update = '-update'
+        _overwrite = '' 
+
+    else: # if not overwrite and path ends with .shp instead
+
+        _update = ''
         _overwrite = ''
 
     # run the final command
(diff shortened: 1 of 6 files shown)


pysqldb3/geospatial.py
```diff
@@ -132,7 +131,7 @@ def write_geo_cmd_query(dbo, query_or_table, is_query = False, schema = ''):
                                 col=col_name, shortened_col = shortened_col
                                 )
             elif dbo.type == MS:
-                results += " , cast([{col}] as date) [{shortened_col}_dt], cast(cast([{col}] as time) as varchar)" \
+                results += " , convert(datetime, [{col}]) [{shortened_col}_dt], convert(varchar, convert(time, [{col}]))" \
```
shostetter: why are you using datetime instead of date?

pysqldb3/tests/helpers.py
```diff
@@ -282,33 +282,6 @@ def clean_up_feature_class():
     # os.remove(os.path.join(fldr, gdb))
     os.rmdir(fldr)
 
-# def set_up_fc_and_shapefile():
```
shostetter, line 285: why are you removing this?

pysqldb3/tests/helpers.py
```diff
@@ -525,4 +498,55 @@ def clean_up_geopackage():
...
+    db_int = data_types_db[0]
+    db_geom = data_types_db[1]
+
+    ### MS SQL SERVER ###
+    # not done in temp table since there is an issue with geometry dtype in SQL Server
+
+    data_types_sql = sql.dfquery(f"""
+                                 
+        drop table if exists [{ms_schema}].[test_default_dtypes];
+        create table [{ms_schema}].[test_default_dtypes] (int_col int, geom geometry);
```
shostetter: most of this can be the same 
create table code can be the same and then cant you use the get table colmns from pysqldb3 to see how the dattypes are coded?

pysqldb3/tests/helpers.py
```diff
@@ -525,4 +498,55 @@ def clean_up_geopackage():
         if os.path.isfile(_fle):
             os.remove(_fle)
 
-    print ('Deleting any existing gpkg')
\ No newline at end of file
+    print ('Deleting any existing gpkg')
+
+
+def identify_default_dtypes(db, sql, ms_schema = ''):
```
shostetter: i think you'd want to just have 1 db connection input and just one dtype output, then call it from the sql and pg as needed

pysqldb3/tests/test_geospatial.py
```diff
@@ -672,6 +648,40 @@ def test_write_gpkg_add_table(self):
         db.drop_table(schema=pg_schema, table=test_reuploaded_table_name + '_2')
         os.remove(os.path.join(FOLDER_PATH, gpkg_name))
 
+
+    def test_write_gpkg_dates_table(self):
+            
+        gpkg_name = 'testgpkg.gpkg'
+
+        db.drop_table(schema = pg_schema, table = test_write_gpkg_table_name)
+        db.drop_table(schema = pg_schema, table = test_reuploaded_table_name)
```
shostetter: this table `test_reuploaded_table_name` isnt used in this test

pysqldb3/tests/test_geospatial.py
```diff
@@ -827,14 +837,15 @@ def teardown_class(cls):
...
+
     def test_write_gpkg_table(self):
         sql.drop_table(schema=ms_schema, table=test_write_gpkg_table_name)
 
         # Add test_table
         sql.query(f"""
-        create table {ms_schema}.{test_write_gpkg_table_name} (test_col1 int, test_col2 int, geom geometry);
-        insert into {ms_schema}.{test_write_gpkg_table_name} VALUES(1, 2, geometry::Point(985831.79200444, 203371.60461367, 2263));
-        insert into {ms_schema}.{test_write_gpkg_table_name} VALUES(3, 4, geometry::Point(985831.79200444, 203371.60461367, 2263));
+        create table {ms_schema}.{test_write_gpkg_table_name} (test_col1 int, test_col2 int, dte datetime, geom geometry);
```
shostetter: isnt this being testing in `test_write_gpkg_dates_table`?

pysqldb3/tests/test_geospatial.py
```diff
@@ -2086,13 +2127,44 @@ def test_write_shp_query(self):
...
+        
+        s.write_geospatial(path=os.path.join(FOLDER_PATH, shp_name), dbo=db, schema=pg_schema, table=test_write_shp_table_name, overwrite = True)
+        
+        # Assert successful
+        assert os.path.isfile(os.path.join(FOLDER_PATH, shp_name))
+
+        # check that Date column is set as a Date type
+        cmd_shp = f'ogrinfo -so -al "{FOLDER_PATH}/{shp_name}" '
+        ogr_response_shp = subprocess.check_output(shlex.split(cmd_shp), stderr=subprocess.STDOUT)   
+        assert 'dt_col_dt: Date' in str(ogr_response_shp), "'dt_col_dt column is not a Date data type when it should be"
```
shostetter, line 2155: should also check that the time date was written?

pysqldb3/tests/test_query_to_geospatial.py
```diff
@@ -859,7 +872,7 @@ def test_query_to_gpkg_data_long(self):
         select
             case when t1.fld2 = t2.fld2 then 1 else 0 end,
             case when left(t1.fld3, 254) = left(t2.fld3, 254) then 1 else 0 end,
-            case when cast(t1.longfld4 as date)=t2.longfld_dt then 1 else 0 end, -- shapefiles cannot store datetimes
+            case when convert(datetime, t1.longfld4)=t2.longfld_dt then 1 else 0 end, -- shapefiles cannot store datetimes
```
shostetter: again why datetime?

shostetter (commented):
1st, thank you for the very clear PR message that was very helpful! 
also i left some minor comments

pysqldb3/tests/helpers.py
cindyatnycdot: it was commented out so I thought I would just clean it up

PR #181, opened 2025-11-18
fix for geopackage default geo name

shostetter:


cindyatnycdot:
how about I just incorporate these changes since there are only 2 small changes into my PR on `rename_geom`?

shostetter:
> how about I just incorporate these changes since there are only 2 small changes into my PR on `rename_geom`?

its just to preserve the history

PR #185, opened 2025-12-03
add sharepoint functions to pysqldb3

Horaceshi624:
add table_to_sharepoint() and sharepoint_to_table() to pysqldb3

note: this now only works for excel/xlsx since mostly files in sharepoints are xlsx

### DIFF
--- a/pysqldb3/pysqldb3.py
+++ b/pysqldb3/pysqldb3.py
@@ -10,7 +10,14 @@ import configparser
 import os
 from .Config import write_config
 import pyarrow.csv as pyarrowcsv
+import shutil
+import tempfile
 import numpy as np
+
+from io import BytesIO
+from office365.sharepoint.files.file import File
+from office365.sharepoint.client_context import ClientContext
+from office365.runtime.auth.user_credential import UserCredential
 
 write_config(confi_path=os.path.dirname(os.path.abspath(__file__)) + "\\config.cfg")
 config = configparser.ConfigParser()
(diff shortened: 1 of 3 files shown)


pysqldb3/pysqldb3.py
```diff
@@ -1079,6 +1086,72 @@ def dataframe_to_table(self, df, table, table_schema=None, schema=None, overwrit
...
+        file_stream.seek(0)
+
+        df = pd.read_excel(file_stream, sheet_name=sheet_name)
+        print(f"Successfully loaded sheet '{sheet_name}' with {len(df)} rows.")
+
+        # Close memory stream
+        file_stream.close()
+
+        # Write DataFrame to DB
+        self.dataframe_to_table(
```
shostetter: i think you should add a check for if the df has more than 1k rows if writes to csv and imports that otherwise it could be slow with large files

pysqldb3/pysqldb3.py
```diff
@@ -2121,6 +2194,94 @@ def table_to_csv(self, table, schema=None, strict=True, output_file=None, open_f
...
+
+        #Export table → CSV
+        self.query_to_csv(
+            query=f'select * from {schema}.{table}',
+            strict=True,
+            output_file=local_csv_path,
+            sep=sep,
+            quote_strings=quote_strings,
+            overwrite=overwrite
+        )
```
shostetter: if you need it to be an excel file why not just write to xlsx from the db rather than going though a csv?

pysqldb3/tests/test_sharepoint_to_table.py
```diff
@@ -0,0 +1,536 @@
...
+            """)
+
+        db.query(f"""
+                insert into {cls.schema}.{cls.src_mixed} values
+                (1, 10, 1.5, 'a', true,  '2024-01-01 10:30:00'),
+                (2, 20, 2.5, 'b', false, '2024-01-02 15:45:00'),
+                (3, null, null, null, null, null);
+            """)
+
+        db.table_to_sharepoint(
```
shostetter: if possible to just write the file to sharepoint without relying on the pysqldb3 functions that would be better

shostetter (commented):
i left a few comments

pysqldb3/pysqldb3.py
Horaceshi624: ok I think I can just use xls to table, seems the easiest

Horaceshi624: oh actually xls to table is also using df to table, I will change to csv to table

PR #186, opened 2025-12-08
Refresh the change to dtypes

cindyatnycdot:
Follow up from last week where I needed to clean up the tests since gpkg didn't always have date errors

#### To-dos:

1.  query the user's data types for geometry (I included integer as well) instead of hardcoding them since they can vary by user and we want the tests to work for all users.
 2.   Add tests for the datetime fields so we know that they are working in PG and MS (this impacts Shp files AND Gpkg/PG)
3.    Edit `write_geo_cmd_query` so that we didn't need an extra line of code to remove columns where the data_type is None. This was happening because MS temp tables don't recognize Geometry fields and would set them to None.

### DIFF
--- a/pysqldb3/geospatial.py
+++ b/pysqldb3/geospatial.py
@@ -78,13 +78,14 @@ def geospatial_tbl_exists(path, geospatial_tbl):
         
     return geo_tbl_exists
 
-def write_geo_cmd_query(dbo, query_or_table, is_query = False, schema = ''):
+def write_geo_cmd_query(dbo, query_or_table, path, is_query = False, schema = ''):
 
     """
     Format the query or table from the database to be written to a Geospatial file.
     Formats columns with special characters and/or datetime columns to be compatible with output file.
     :param dbo: Database connection
     :param query_or_table: The specific query or table name to be written to a Geospatial file
+    :param path: Output file path to check if it is a Shp file
     :param is_query (bool): Boolean for whether or not it is a query being written to Geospatial file.
                             Defaults to False (table); if it's query, set to True.
     :param schema: Optional schema name; used only if the command is calling a table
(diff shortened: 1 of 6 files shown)


pysqldb3/geospatial.py
```diff
@@ -109,8 +110,7 @@ def write_geo_cmd_query(dbo, query_or_table, is_query = False, schema = ''):
     columns_with_brackets = [left_bracket + c[0] + right_bracket for c in columns_with_types]
 
     # identify date columns if they are in the intended output
-    dt_col_names = [c for c in columns_with_types if c[1] is not None] ## TODO - what is this doing/is it needed?
-    dt_col_names = [c[0] for c in dt_col_names if (('datetime' in c[1]) | ('timestamp' in c[1]))] # after, check if there is datetime / timestamp
+    dt_col_names = [c[0] for c in columns_with_types if (('datetime' in c[1]) | ('timestamp' in c[1]))] # after, check if there is datetime / timestamp
 
     # if there are no date columns, we can query immediately
```
shostetter, line 142: i think the datetime logic should live in its own function that can be called here, then you can call it only f the output is shp or if output is repackage and source is PG

pysqldb3/tests/test_geospatial.py
```diff
@@ -2086,12 +2065,35 @@ def test_write_shp_query(self):
...
+
+        s.write_geospatial(path=os.path.join(FOLDER_PATH, shp_name), dbo=db, schema=pg_schema, table=test_write_shp_table_name, overwrite = True)
+
+        # Assert successful
+        assert os.path.isfile(os.path.join(FOLDER_PATH, shp_name))
+
+        # check that Date column is set as a Date type
+        cmd_shp = f'ogrinfo -so -al "{FOLDER_PATH}/{shp_name}" '
+        ogr_response_shp = subprocess.check_output(shlex.split(cmd_shp), stderr=subprocess.STDOUT)   
+        assert 'dt_col_dt: Date' in str(ogr_response_shp), "'dt_col_dt column is not a Date data type when it should be"
```
shostetter: shouldnt you test all of the dt_ columns? and check for the time 1/2 of the data?

shostetter (commented):
left some minor comments - the reorg one in `write_geo_cmd_query` i think we can just add as a todo in the next pairing - so really just a question about the test

pysqldb3/tests/test_geospatial.py
cindyatnycdot: And check that it is a Varchar column? Since we know it doesn't become a Time

cindyatnycdot:
> left some minor comments - the reorg one in `write_geo_cmd_query` i think we can just add as a todo in the next pairing - so really just a question about the test

I made the date formatting function since it was mostly just moving a portion of the code. We can review it during the pair programming if you want

shostetter (approved):

PR #189, opened 2025-12-23
180 create multi row header workflow for xls/csv to table

### LINKED ISSUE #180: create multi-row header workflow for xls to table
sample `C:\Users\SHostetter\Desktop\GIT\MiscRequests\Internal\TPM\Requests\TIGER_Grant-crash_req\FINAL NY Vision Zero Performance Measures_Bas.xlsx` 
```
df = pd.read_excel('.\FINAL NY Vision Zero Performance Measures_Bas.xlsx', header=[7,8])
df.head()

df.columns = df.columns.map('-'.join)
df.columns
```

this should be handled on import under the hood

shostetter:


pysqldb3/pysqldb3.py
```diff
@@ -1146,12 +1151,13 @@ def contains_long_columns(df2):
             try:
                 temp_file = os.path.dirname(input_file)+'\\'f'_temp_data_{datetime.datetime.now().strftime("%Y%m%d%H%M%S")}.csv'
 
-
+                # df.to_csv(temp_file, chunksize=10 ** 15, index=False)
```
cindyatnycdot, line 1155: Should this be deleted?

pysqldb3/tests/helpers.py
```diff
@@ -396,7 +442,9 @@ def set_up_xls():
 
     test_df2 = pd.DataFrame({'a': {0: 1, 1: 2}, 'b': {0: 3, 1: 4}, 'Unnamed: 0': {0: 0, 1: 1}})
 
-    test_df2.to_excel(os.path.join(DIR, 'test_xls_with_sheet.xls'), sheet_name='AnotherSheet', index=False)
+    # test_df2.to_excel(os.path.join(DIR, 'test_xls_with_sheet.xls'), sheet_name='AnotherSheet', index=False)
```
cindyatnycdot: Do you want to delete this since you ended up commenting it out?

pysqldb3/tests/helpers.py
```diff
@@ -387,7 +431,9 @@ def set_up_xls():
         clean_up_file(xls_file1)
 
     test_df1 = pd.DataFrame({'a': {0: 1, 1: 2, 2:3}, 'b': {0: 3, 1: 4, 2:5}, 'Unnamed: 0': {0: 0, 1: 1, 2:6}})
-    test_df1.to_excel(os.path.join(DIR, 'test_xls.xls'), index=False)
+    # test_df1.to_excel(os.path.join(DIR, 'test_xls.xls'), index=False)
```
cindyatnycdot: Do you want to delete this since you ended up commenting it out?

pysqldb3/tests/test_dbconnect.py
```diff
@@ -453,7 +453,8 @@ def test_excel_to_table_logging(self):
         assert before_drop_working_log_length == 0
 
         df = pd.DataFrame({"a": [1, 2], "b": [3, 4]})
-        df.to_excel(fp)
+        # df.to_excel(fp)
+        helpers.write_xls_from_df(fp, df)
```
cindyatnycdot, line 457: I think the only thing missing from the `test_dbconnect.py` file is to clean up the files that are created as part of `helpers.set_up_xls()` in row 445 ( `test_xls.xls` and `test_xls_with_sheet.xls` files)

pysqldb3/pysqldb3.py
```diff
@@ -1651,7 +1657,9 @@ def xls_to_table(self, input_file=None, sheet_name=0, overwrite=False, schema=No
 
         try:
             df = pd.read_excel(input_file, sheet_name=sheet_name, **kwargs)
-
+            if 'header' in kwargs:
+                # todo clean out unnamed sub headers
+                df.columns = df.columns.map('_'.join)
```
cindyatnycdot: same comment as above about `header` argument

pysqldb3/pysqldb3.py
```diff
@@ -1133,6 +1133,11 @@ def contains_long_columns(df2):
 
         if 'ogc_fid' in df.columns:
             df = df.drop('ogc_fid', 1)
+        # Multi-row headers?
+        if 'header' in kwargs:
+            df.columns = df.columns.map('_'.join)
```
cindyatnycdot: You might need to change the if statement so that if the header argument contains a list of integers, that's the only time when `df.columns.map('_'.join)` gets executed.

technically, under `pandas.read_csv()` the user has several input options for the `header` argument (the user can also input 'infer', None, or a single integer).  The `csv_to_table` function doesn't return what I'm expecting if I put `header = 0`  and it errors out when `header = None`, which a user of the function might try to do (since these are valid inputs in `pandas.read_csv`).

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html

shostetter: ah this is a good catch, thank you!

cindyatnycdot (approved):

PR #195, opened 2026-02-20
Create tests for a zip file with a subfolder

### LINKED ISSUE #196: Add tests for geospatial zip files where the file is in a subfolder

cindyatnycdot:
Context for this PR is that Seth asked Cindy to create an additional test for the scenario that a zip file has a subfolder.

### DIFF
--- a/pysqldb3/tests/test_geospatial.py
+++ b/pysqldb3/tests/test_geospatial.py
@@ -1604,6 +1604,47 @@ class TestReadShpPG:
         assert len(diff_df) == 1
         assert int(diff_df.iloc[0]['distance']) == 0
 
+    def test_read_shp_subzip(self):
+
+        fp = FOLDER_PATH + '/subfolder.zip/extra_subfolder/'
+        shp_name = "test.shp"
+
+        # Make sure table doesn't alredy exist
+        db.drop_table(pg_schema, test_read_shp_table_name)
+        assert not db.table_exists(schema=pg_schema, table=test_read_shp_table_name)
+
+        # Assert successful
+        db.drop_table(schema=pg_schema, table=test_read_shp_table_name)
+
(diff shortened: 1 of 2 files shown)


pysqldb3/tests/test_geospatial.py
```diff
@@ -1604,6 +1604,47 @@ def test_read_shp_zip(self):
...
+
+        fp = FOLDER_PATH + '/subfolder.zip/extra_subfolder/'
+        shp_name = "test.shp"
+
+        # Make sure table doesn't alredy exist
+        db.drop_table(pg_schema, test_read_shp_table_name)
+        assert not db.table_exists(schema=pg_schema, table=test_read_shp_table_name)
+
+        # Assert successful
+        db.drop_table(schema=pg_schema, table=test_read_shp_table_name)
```
shostetter: I think this is duplicative of line 1617

pysqldb3/tests/test_geospatial.py
```diff
@@ -1888,6 +1928,45 @@ def test_read_shp_zip_2(self):
         assert len(diff_df) == 1
         assert int(diff_df.iloc[0]['distance']) == 0
 
+    def test_read_shp_subzip(self):
```
shostetter, line 2010: isnt this the same as the test on 1607?

shostetter (commented):
few minor questions, but i think you need to add tests for zipped with subfolders feature class and geopackage files too

pysqldb3/tests/test_geospatial.py
cindyatnycdot: Yes, but this is for MS SQL. 1607 is for Postgresql.

shostetter (approved):

PR #200, opened 2026-03-31
Simplify geospatial_convert inputs

cindyatnycdot:


pysqldb3/util.py
```diff
@@ -356,26 +356,40 @@ def print_cmd_string(password_list, cmd_string):
...
     if '.zip' in path.lower():
         path = '/vsizip/' + path
 
-    if file_name:
-        return path, file_name
+    return path
 
-    file_name = os.path.basename(path)
-    path = path.replace(file_name, '')
+def parse_file_path(path = None, file_name = None):
```
shostetter: proposed simplification 

```
def parse_file_path(path=None, file_name=None):
    """
    Parse the file name to separate the path and file name.
    :param path: Optional file path
    :param file_name: Optional file name
    """
    # if no file name is given, then we parse the path
    if not file_name:
        file_name = os.path.basename(path)
        path = path.replace(file_name, '')
    # if not path is given, parse the filename
    if not path:
        path = file_name.replace(os.path.basename(file_name), '')
        file_name = os.path.basename(file_name)
    # combine 
    full_path = os.path.join(path, file_name)
    return full_path, path, file_name
```

pysqldb3/util.py
```diff
@@ -684,28 +698,28 @@ def retrieve_gpkg_tbl_names(dbo, full_path):
 
     return gpkg_tbl_names
 
-def convert_cmd(input_full_path, output_full_path, gpkg_tbl = None, _update = None, _overwrite = None, feature_class = None):
+def convert_cmd(input_full_path, output_full_path, _update = None, _overwrite = None, input_table = None, output_table = None):
```
shostetter, line 709: you should update the doc strings to describe the inputs

pysqldb3/pysqldb3.py
```diff
@@ -2041,7 +2041,7 @@ def query_to_gpkg(self, query, gpkg_tbl, path, overwrite = True, cmd=None,
         write_geospatial(dbo = self, path = path, query = query, gpkg_tbl = gpkg_tbl,
                             overwrite = overwrite, cmd = cmd, gdal_data_loc = gdal_data_loc, print_cmd = print_cmd, srid = srid)
 
-    def shp_to_table(self, path,  shp_name = None, table=None, schema=None, feature_class = None, cmd=None,
+    def shp_to_table(self, path,  shp_name = None, table=None, schema=None, feature_class = None,
```
shostetter: why did you take the cmd option out?

pysqldb3/geospatial.py
```diff
@@ -279,82 +279,98 @@ def write_geospatial(dbo, path,  table = None, schema = '', query = None, gpkg_t
...
-    
-        # if gpkg exists and overwrite was not explicitly called
-        if geospatial_exists(path = export_path) and not geospatial_tbl_exists(path = export_path, geospatial_tbl = gpkg_tbl) and overwrite == False:
-            _update = '-update' # then add the table to the gpkg
+    if '.shp' in input_path:
+        input_table = ''
+    else:
+        # gpkg or gdb
+        input_table = os.path.basename(input_path)
+        input_path = os.path.dirname(input_path)
```
shostetter: cant you use `parse_file_path`

pysqldb3/geospatial.py
```diff
@@ -279,82 +279,98 @@ def write_geospatial(dbo, path,  table = None, schema = '', query = None, gpkg_t
...
-            print("The table name to be copied to the geopackage already exists. Either change to Overwrite = True or check the name of the table to be copied.")
-            exit # stop process so user can fix
-
+    ### create the output file path ###
+    if '.shp' in output_file:
+        output_table = ''
+        
+        if not '/' in output_file and not '\\' in output_file:
+            # if the output_file has no path, set it to the input_path
+            output_path = os.path.join(os.path.dirname(input_path), output_file)
```
shostetter: all of teh input / output files/paths should be set at the beginging of the function so you dont need to keep parsing it

pysqldb3/geospatial.py
```diff
@@ -279,82 +279,98 @@ def write_geospatial(dbo, path,  table = None, schema = '', query = None, gpkg_t
...
+    # assert the OUTPUT file formats are correct
+    if output_file:
+        assert output_file.endswith('.shp') or any(x in output_file for x in ['.gpkg/', '.gpkg\\']), \
+            "The output file must end with .shp or .gpkg/[gpkg_tbl]. Cannot create new Geodatabase via GDAL"
 
-    # if no gpkg_tbl name given and we convert a shp file, name the table consistent with the shapefile
-    if not gpkg_tbl and input_full_path.endswith('.shp'):
-        gpkg_tbl = input_file.replace('.shp', '')
+    # if the file happens to be a zip file, this will help us read it
+    input_path = add_zip_to_geo_path(path)
```
shostetter, line 311: lines 308-353 are just checking the input/output formatting is correct? if so move it to its own function

shostetter (commented):
left some comments - didnt look at the tests

pysqldb3/pysqldb3.py
cindyatnycdot: The `cmd` arugment has never been meaningfully used in the `shp_to_table` function so I just removed it. Even in the pysqldb3 `main` branch, the `read_shp` function in production does not actually use the `cmd` argument (even though `shp_to_table` accepts it as an argument), and my updated function `upload_geospatial` doesn't use it since I modeled it after that one.

pysqldb3/geospatial.py
cindyatnycdot: made a slight change to the function so that it would work for these

pysqldb3/geospatial.py
cindyatnycdot: I thought this was already the beginning of the function. It starts off with some `assert` checks to make sure the inputs are valid, and then it goes into parsing the `input_path`, `input_table`, `output_path`, `output_table` values. I can move the `if overwrite` section so it's after the parsing.

pysqldb3/geospatial.py
cindyatnycdot: Moved to a function in `utils` now

pysqldb3/tests/test_util.py
```diff
@@ -160,3 +161,131 @@ def test_wkt(self):
...
+        file_path = 'E:\\RIS\\Staff Folders\\Seth H\\DATA\\OUTPUT_DATA\\test_zipfile.zip\\subfolder\\test_shapefile.shp'
+
+        new_path = add_zip_to_geo_path(file_path)
+        folder_dir, file_name, table_name = parse_file_path(new_path)
+
+        assert folder_dir == '/vsizip/E:\\RIS\\Staff Folders\\Seth H\\DATA\\OUTPUT_DATA\\test_zipfile.zip\\subfolder'
+        assert file_name == 'test_shapefile.shp'
+        assert table_name == ''
+
+class TestParseFilePathGdb:
```
shostetter, line 216: do we need to add a gdb in a zip file?

pysqldb3/tests/test_util.py
```diff
@@ -160,3 +161,131 @@ def test_wkt(self):
...
+        assert table_name == ''
+
+    def test_input_types_geopakage_zip(self):
+        
+        file_path = 'E:\\RIS\\Staff Folders\\Seth H\\DATA\\OUTPUT_DATA\\zip_file_name.zip\\test_geodb.gpkg\\test_geopackage_table'
+
+        # simple path and file
+        new_path = add_zip_to_geo_path(file_path)
+        file_dir, file_name, table_name = parse_file_path(new_path)
+        assert file_dir == '/vsizip/E:\\RIS\\Staff Folders\\Seth H\\DATA\\OUTPUT_DATA\\zip_file_name.zip'
```
shostetter, line 331: should the zip be in the directory? i think youd want the folder not the zip file...? Aren't you using the folder to write back to, or would you  write it back to the zip file, and if so are there tests for that?

pysqldb3/geospatial.py
shostetter: is this still needed? why would only shp need to be joined?

shostetter (commented):
left a few questions

pysqldb3/tests/test_util.py
cindyatnycdot: I treated files ending with `.gdb/.gpkg/.shp` as the 'file', but I realized from the tests you created that you consider `.gpkg/.gdb` as part of the 'directory'. I think they could be classified either way; do you think it's fine to keep them as files? The end user does not need to know what portion I considered the file directory vs the file, so I don't think this will be a problem.

Right now, any `zip` functionality is for _reading_ geospatial files, but not writing them.

pysqldb3/tests/test_util.py
cindyatnycdot: I will add a test

pysqldb3/tests/test_util.py
cindyatnycdot: I've added something that checks if a file is writing back to a zip, and if so, it will save the file higher in the directory where it is no longer in a zip file. It will also print a message to tell user the new output file directory  if they did not provide an output folder directory and the original input was in a zip.

pysqldb3/geospatial.py
cindyatnycdot: I've now altered the logic in `parse_file_path` slightly so that the if statement isn't needed

cindyatnycdot (commented):
I made some small changes based on your comments.

shostetter (approved):
looks good!

PR #211, opened 2026-06-11
209 add cross tab query

shostetter:


pysqldb3/pysqldb3.py
```diff
@@ -943,6 +941,54 @@ def print_last_query(self):
...
 
+    def crosstab_query(self, qry, row_columns, column_column, count_column=None, sum_column=None, df=False, **kwargs):
+        """
+        Creates crosstab query from input query. Selects from the input query data and aggregates on the row_columns,
+        creates new columns for each unique value in the column_column field and then sums or counts the values in sum_column or count_column.
+        :param qry: Query string to use as base data
+        :param row_columns: List of column names to use for row aggregation
+        :param column_column: Column to use for the cross, each value will become a new column in the output
+        :param count_column: Column whose values will be counted
+        :param sum_column: Column whose values will be summed
```
cvilter, line 952: Instead of having count_column and sum_column, have you considered having aggregate_column and aggregate_function? For ex 
```aggregate_column = 'vehicle_id', aggregate_function='count'```
This way both arguments are used every time and there isn't ambiguity about what aggregate function is being used (e.g. the reader of a query like the sample in the Readme doesn't have to trace back to the documentation whether count_column or sum_column comes first in the argument list). It might make it easier to support more aggregate functions, too.

File: pysqldb3/pysqldb3.py [hunk 2]
```diff
@@ -943,6 +941,54 @@ def print_last_query(self):
...
+                cols += (f"""sum(distinct case when {column_column} = '{_}' then {sum_column} else null end) as "{_}",""")
+
+        # build full query string
+        _qry = f"select {str(row_columns)[1:-1].replace("'","")}, {cols[:-1]} from ({qry}) t group by {str(row_columns)[1:-1].replace("'","")}"
+
+        # pass new crosstab query to query or df query to get results
+        if df:
+            return self.dfquery(_qry, **kwargs)
+        else:
+            return self.query(_qry, kwargs)
```
cvilter: Is there a reason for the `**` on one and not the other?

pysqldb3/pysqldb3.py
cvilter: lines 977 and 981 would be combined into one, taking the aggregate function from the input

pysqldb3/pysqldb3.py
```diff
@@ -943,6 +941,54 @@ def print_last_query(self):
...
+        if count_column:
+            for _ in col_values:
+                cols += (f"""count(distinct case when {column_column} = '{_}' then {count_column} else null end) as "{_}",""")
+
+        if sum_column:
+            for _ in col_values:
+                cols += (f"""sum(distinct case when {column_column} = '{_}' then {sum_column} else null end) as "{_}",""")
+
+        # build full query string
+        _qry = f"select {str(row_columns)[1:-1].replace("'","")}, {cols[:-1]} from ({qry}) t group by {str(row_columns)[1:-1].replace("'","")}"
```
cvilter, line 984: I kept getting an error when I tried to run this as-is (prob bc I'm running 3.11), which I think comes down to confusion around `"'",""`. What resolved it for me was replacing line 984 with these two lines:

```
cleaned_row_cols= str(row_columns)[1:-1].replace("'", "")
_qry = f"select {cleaned_row_cols}, {cols[:-1]} from ({qry}) t group by {cleaned_row_cols}"
```

cvilter (commented):
I'm running into this error when I try to run the tests - can't immediately figure out why - lmk if you have insights 

Image

pysqldb3/pysqldb3.py
shostetter: the sum/count columns are keyword args so the order doesnt matter you pass in with the keyword. The agg function idea is a good one, but i worry it opens it up to way more complexity and would be both more fragile and harder to test, might be worth thinking about as another kwarg that could be used for other agg types

pysqldb3/pysqldb3.py
shostetter: so you just had to split it into 2 parts?

pysqldb3/pysqldb3.py
shostetter: ixed

pysqldb3/pysqldb3.py
cvilter: Yes, python 11 gets upset about the quotation marks otherwise aka breaks

cvilter (approved):
All tests pass. The change I recommended for line 984 is worth making for backwards compatibility for those of us who don't have python 3.14; no add'l comments

cvilter (approved):
All tests pass

PR #216, opened 2026-07-30
edit the SERVER param for the Azure connection

cindyatnycdot:
I couldn't get the azure connection to work until I made this change, because I got an error that `pymssql` did not recognize the SERVER parameter. Is it working for you? Not sure if this is a version issue....


shostetter (approved):
works both ways for me

